PR #2, opened 2022-08-10
EVG-17439: Create base evergreen.yml config

minnakt:
EVG-17439

### Description 
This PR introduces a basic `.evergreen.yml` config for Parsley. It currently only tests two things, `compile` and `type_check`. We will add more tasks to the config as we introduce Storybook, Cypress, etc.

Tagging Sanjog because the ticket says to do so.

### Testing 
* This patch should pass!

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -19,7 +19,7 @@ functions:
     params:
       working_dir: parsley
       script: |
-        export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
+        export PATH=${node_path}
         yarn build:local
 
   yarn-install:
@@ -27,7 +27,7 @@ functions:
     params:
       working_dir: parsley
       script: |
-        export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
+        export PATH=${node_path}
         yarn
 
   yarn-tsc:
@@ -35,7 +35,7 @@ functions:
     params:
       working_dir: parsley
       script: |
-        export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
+        export PATH=${node_path}
         yarn check-types
 
 #######################################
@@ -58,7 +58,7 @@ buildvariants:
   - name: ubuntu1604
     display_name: Ubuntu 16.04
     run_on:
-    - ubuntu1604-large
+    - ubuntu1604-small
     tasks:
     - name: compile
     - name: type_check
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -14,14 +14,6 @@ functions:
     params:
       directory: parsley
 
-  yarn-build:
-    command: shell.exec
-    params:
-      working_dir: parsley
-      script: |
-        export PATH=${node_path}
-        yarn build:local
-
   yarn-install:
     command: shell.exec
     params:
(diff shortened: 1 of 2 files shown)


.evergreen.yml
```diff
@@ -1,19 +1,64 @@
...
+      script: |
+        export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
+        yarn build:local
+
+  yarn-install:
+    command: shell.exec
+    params:
+      working_dir: parsley
+      script: |
+        export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
```
sophstad, line 30: Since we're using Node v16.13 we could add an `engines` section to `package.json` like we have [in Spruce](https://github.com/evergreen-ci/spruce/blob/4fd10c45ffed5e8551f73bef326efc6cebb3eb68/package.json#L233-L235)

.evergreen.yml
```diff
@@ -1,19 +1,64 @@
...
+    type: setup
+    params:
+      directory: parsley
+
+  yarn-build:
+    command: shell.exec
+    params:
+      working_dir: parsley
+      script: |
+        export PATH=/opt/nodejs/node-v16.13.0-linux-x64/bin:$PATH
```
sophstad, line 22: Could we add this line either to `get-project` or as its own function in order to avoid repetition? Will be easier to upgrade our version of Node in the future :)

.evergreen.yml
```diff
@@ -1,19 +1,64 @@
...
+    - func: get-project
+    - func: yarn-install
+    - func: yarn-tsc
 
 buildvariants:
   - name: ubuntu1604
     display_name: Ubuntu 16.04
     run_on:
-      - ubuntu1604-small
+    - ubuntu1604-large
```
sophstad, line 61: We could maybe start off using a small distro until our Cypress testing gets significant (save dat money 🤑)

sophstad (changes requested):

.evergreen.yml
minnakt: It was not very pleased when I tried to move these scripts out (my [patch](https://spruce.mongodb.com/version/62f414eee3c3311893d459d8/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)), so I went for an expansion instead 🤔

sophstad: So actually it looks like the problem with your patch is that `script` should be a field within the `params` field (i.e. indented under it). YAML syntax is so hard 😭. I ran a successful patch with that approach [here](https://spruce.mongodb.com/version/62f664810305b95854d1f747/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). Mind going with that instead of the variable? I feel like it's a little simpler.

minnakt: I tried again but I couldn't get it to work 🥲 ([my patch](https://spruce.mongodb.com/version/62f667039ccd4e2d11fa5657/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))

I also tried it as a separate command in [this patch](https://spruce.mongodb.com/version/62f422d1e3c3311893d47b61/changes?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

I think the patch you linked may have passed because each of the commands still have individual export statements like this `export PATH=${node_path}`, or I am just very bad at copy pasting

minnakt:
lmk if I did that right 😅

sophstad (approved):
minna, it looks incredible!!

minnakt:
Commenting on why we weren't able to extract the path script out for future readers: we suspect that `shell.exec` starts with a fresh environment, which means that we need to set the `PATH` variable every time.

PR #3, opened 2022-08-11
EVG-17440: Configure ESLint and Prettier

minnakt:
EVG-17440

### Description 
This PR configures ESLint and Prettier for Parsley.

Notable changes from Spruce:
* I omitted the `.prettierrc` file from Spruce. Everything in that file is actually set to the defaults, so the file doesn't really add anything to the default Prettier config.
* Organized and grouped rules together so that it's easier to see what rules are being disabled.

I'm on the fence about it, but I used the `airbnb` config because it has some stricter rules that are useful. We don't technically have to use it though. We could instead use `eslint:all / react:all` instead of `eslint:recommended / react:recommended` and then disable the rules we aren't interested in one by one.

### Interesting Reading
It seems that ESLint eventually wants to move to a new file format called `eslint.config.js`. They are envisioning a single config file in one location for configuring ESLint. However, it's still an experimental feature and use-at-your-own-risk.

Here's an [article](https://eslint.org/blog/2022/08/new-config-system-part-2/) describing their new config system. Here's the [documentation](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new) of `eslint.config.js`.

### Testing 
* Added lint task to `.evergreen.yml`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,108 @@
+const ERROR = "error";
+const WARN = "warn";
+const OFF = "off";
+
+const errorIfStrict = process.env.STRICT ? ERROR : WARN;
+
+module.exports = {
+  env: {
+    browser: true,
+    es6: true,
+    node: true,
+    // TODO EVG-17441: add Jest here.
+  },
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true,
+    },
+    ecmaVersion: "latest",
+    sourceType: "module",
+  },
+  extends: [
+    "eslint:recommended",
+    "plugin:import/recommended",
+    "plugin:react/recommended",
+    "plugin:prettier/recommended", // Note: prettier must ALWAYS be the last extension.
+  ],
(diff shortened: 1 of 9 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,121 @@
...
+      jsx: true,
+    },
+    ecmaVersion: "latest",
+    sourceType: "module",
+  },
+  extends: [
+    "eslint:recommended",
+    "plugin:import/recommended",
+    "plugin:react/recommended",
+    "airbnb", // Airbnb includes some helpful rules for ESLint and React that aren't covered by recommended.
```
sophstad, line 26: I'm down to give Airbnb rules a try in Parsley -- if we don't like them we can disable them, and if we do we can add them to Spruce!

.evergreen.yml
```diff
@@ -21,7 +21,7 @@ functions:
       script: |
         export PATH=${node_path}
         yarn
-
+
```
sophstad, line 24: [nit] Remove

sophstad (approved):
sweet

PR #4, opened 2022-08-12
EVG-17441: Configure Jest

minnakt:
EVG-17441

### Description 
This PR configures Jest for Parsley.

Notable changes from Spruce:
* In Spruce we use `babel/jest` to transform our React files into a format understandable by Jest. I decided to try using [`ts-jest`](https://github.com/kulshekhar/ts-jest) which is a library that is acknowledged on the official [Jest docs](https://jestjs.io/docs/getting-started#via-ts-jest). `ts-jest` handles transformations like `babel/jest` and has the ability to type check tests as they are run, which Babel cannot do. I'm okay with changing it back if that's preferred though.
* In Spruce we have two transform scripts`cssTransform.js` for CSS imports and `fileTransform.js` which I think is for SVG imports(?). I left them out because they're quite old and I'm not sure how necessary they are. I will be revisiting the Jest config when installing Emotion and LG, so if they are needed I can just put them in that PR.
* In Spruce we had a script for running Jest in `scripts/test.js`. I think the important part of this script is that it is used to configure the output directory for the Jest tests. We can actually do that in `jest.config.js`, so I omitted this file. It also seems a little outdated, considering it links to an issue from nearly 4 years ago and was written while we were using Node 12 on CI.

### Testing 
* Added tests for `App.test.tsx` that should pass on the new `test` task


package.json
```diff
@@ -12,13 +12,22 @@
     "eslint:strict": "STRICT=1 eslint -c src/.eslintrc.js 'src/**/*.ts?(x)'",
     "prettier": "prettier --write",
     "preview": "vite preview",
-    "start": "vite"
+    "start": "vite",
+    "test:exact": "jest --",
```
sophstad, line 16: It looks like the [syntax](https://jestjs.io/docs/cli#running-from-the-command-line) for running a specific test is just `jest <path/to/test>`. As such, I think we can use `yarn test <path/to/test>` to run a specific test, no `test:exact` needed — I added another test file and tried locally and this seems to work!

package.json
```diff
@@ -12,13 +12,22 @@
...
+    "test:watch": "jest --watch  --verbose",
+    "test": "jest --watchAll=false"
   },
   "dependencies": {
     "react": "^18.2.0",
-    "react-dom": "^18.2.0"
+    "react-dom": "^18.2.0",
```
sophstad, line 22: Any reason we shouldn't pin these both to `18.2.0`?

src/test_utils/index.tsx
```diff
@@ -0,0 +1,94 @@
+import {
```
sophstad, line 1: I hate to have such a duplicate of a Spruce file 😩 maybe down the line we can think about how best to share this code. But looks good for now!

sophstad (changes requested):

package.json
minnakt: no reason, I think the Vite project was just generated like that with the `^` dependencies! I pinned them 😈

src/test_utils/index.tsx
minnakt: 😔😭👍

sophstad (approved):
Nice work! I think `ts-jest` looks like the right move, too 👍😄

PR #5, opened 2022-08-15
EVG-17442: Configure Cypress

minnakt:
EVG-17442

### Description 
This PR configures Cypress for Parsley. 

Notable changes from Spruce:
* Not all of the custom commands from Spruce are included in `commands.ts`, but they can always be added later very easily.
* We have a script called `prodServer.js` on Spruce. To be honest I'm not exactly sure, but I think it serves the static build files to localhost so that we run Cypress on a build more representative of production. I think this also makes Cypress run faster. From what I can tell, using `yarn build` and then `yarn preview` does the same thing ([Vite docs](https://vitejs.dev/guide/static-deploy.html)) so I just went that route. (But let me know if that seems wrong)
* I haven't set up the Cypress Dashboard, but I'm going to kick that stuff over to a later ticket.

### Testing 
- Added Cypress task to `.evergreen.yml`


.evergreen.yml
```diff
@@ -77,12 +138,22 @@ tasks:
   - name: lint
     commands:
     - func: yarn-eslint
-
+
```
sophstad, line 141:
```suggestion

```

package.json
```diff
@@ -7,14 +7,17 @@
     "build": "tsc && vite build",
     "build:local": "yarn build",
     "check-types": "tsc -p tsconfig.json --noEmit",
+    "cy:open": "cypress open",
+    "cy:run": "cypress run",
+    "cy:test": "cypress run --spec",
```
sophstad, line 12: Since `--spec` needs an argument, it feels more transparent to just use `yarn cy:run --spec <spec>` instead of having a separate command
```suggestion
```

package.json
```diff
@@ -7,14 +7,17 @@
...
     "dev": "vite",
     "eslint:fix": "STRICT=1 eslint -c src/.eslintrc.js --fix 'src/**/*.ts?(x)'",
     "eslint:strict": "STRICT=1 eslint -c src/.eslintrc.js 'src/**/*.ts?(x)'",
     "prettier": "prettier --write",
     "preview": "vite preview",
     "start": "vite",
-    "test:watch": "jest --watch  --verbose",
-    "test": "jest --watchAll=false"
+    "test": "jest --watchAll=false",
+    "test:watch": "jest --watch  --verbose"
```
sophstad, line 20: good improvement 😎

sophstad (changes requested):

.evergreen.yml
minnakt: why does my YAML plugin always fail me 🧍

minnakt:
I moved the e2e task to `ubuntu1604-large` because we will probably need that in the future.

sophstad (approved):
> I moved the e2e task to ubuntu1604-large because we will probably need that in the future.

Nice idea splitting this into two build variants! LGTM

PR #6, opened 2022-08-16
EVG-17443: Configure Storybook

minnakt:
EVG-17443

### Description 
This PR configures Storybook and storyshots for Parsley. The stories included in this PR are all auto-generated from installing Storybook. I temporarily unignored CSS files in the `.gitignore` for the sake of testing storyshots, but I will re-ignore them in the PR for Emotion/Leafygreen.

From what I have read it seems like `@storybook/addon-storyshots` does not have strong support because the maintainers are planning to create a new addon to replace it. It seems like it works best with Jest 26 and before, but I don't want to downgrade our Jest version.

Notable changes from Spruce:
* I could not get storyshots working with `.storybook/main.ts`, only with `.storybook/main.js` 😭
* In Spruce `.storybook/main.ts`, we have some code to manipulate config plugins because of issues that arise in production. However the GitHub issue linked in that file has since been closed with a comment saying that it should be resolved in `@storybook/builder-vite v0.2.1`. For that reason I left this code out.
* In `package.json`, I had to add a resolutions field for `react-test-renderer` because apparently React 18 does not play well with storyshots without this. I wish I could link the GitHub issue in the `package.json` but you can't leave any comments in the file, so I left one in `.storybook/main.js` instead. 🥲 The relationship between storyshots and `react-test-renderer` is briefly described in this  [README](https://github.com/infinitered/addon-storyshots/blob/master/README.md).

### Testing 
- Added Storybook and storyshots task in `.evergreen.yml`


src/stories/Button.stories.tsx
```diff
@@ -0,0 +1,40 @@
+import { ComponentMeta, ComponentStory } from "@storybook/react";
```
sophstad, line 1: Do you anticipate deleting the `stories/` directory once we start development?

.evergreen.yml
```diff
@@ -154,6 +168,15 @@ tasks:
     - func: attach-cypress-screenshots
     - func: attach-cypress-videos
 
+  - name: storybook
+    commands:
+    - func: yarn-storybook
```
sophstad, line 173: Would it be possible to add a step to attach Storybook files to the task? We do this [in Spruce](https://github.com/evergreen-ci/spruce/blob/92d42b3794307d0f034944a7a806898342cc720a/.evergreen.yml#L234-L246)!

sophstad (changes requested):

src/stories/Button.stories.tsx
minnakt: Yep, I will delete the entire directory! In my Emotion/LG PR, I will add a story for a custom component using Emotion and LG and make sure that everything still works

.evergreen.yml
```diff
@@ -47,6 +47,20 @@ functions:
...
+    params:
+      aws_key: ${aws_key}
+      aws_secret: ${aws_secret}
+      local_files_include_filter:
+        ["parsley/storybook-static/*"]
+      remote_file: parsley/${task_id}/storybook/
+      bucket: mciuploads
+      content_type: text/html
+      permissions: public-read
+      display_name: storybook_
```
sophstad, line 63: Hmm, the ultimate goal of these files is that clicking on `<display_name>index.html` in the Files tab should link to a working Storybook build, but clicking `storybook_index.html` here points to a broken site. Is it possible that using `storybook-` here like in Spruce would fix it?

sophstad (changes requested):

sophstad (approved):
Wow great job getting this working :')

minnakt:
Commenting for future readers

Storybook on production is broken for Spruce, but should be working on Parsley. The following are the changes needed for it to work:
* In `.storybook/main.js`, `base` needs to be set to `"./"` in `viteFinal`. Without this, the path to the `/assets` folder is an absolute path which means Storybook can never find the corresponding `/assets` folder on production.
* `s3.put` doesn't preserve folder structure, so I added commands to put the appropriate files in the `storybook-static/assets` directory in the S3 bucket. Note that I had to use a separate command for each file type in `/assets` because S3 was not happy unless the MIME type was exact... 🤔 Possibly [EVG-17334](https://jira.mongodb.org/browse/EVG-17334) will alleviate this in the future!

PR #8, opened 2022-08-19
EVG-17445, EVG-17464: Configure Emotion and LG

minnakt:
EVG-17445 and EVG-17464

### Description 
This PR configures Emotion and LG for Parsley. I created a very simple component using Emotion and LG in `components/Searchbar` and tested that it works with Storybook and Jest.

Notable changes:
* Deleted the `/src/stories` folder which had been auto-generated by Storybook.
* I was able to delete the Jest setup script in `config/jest/setup.js` by changing `jest.config.js` to import the necessary library.
* Leafygreen introduced a function [`createUniqueClassName`](https://github.com/mongodb/leafygreen-ui/blob/main/packages/lib/src/createUniqueClassName/index.ts#L20) which generates unique classnames at run time. This is problematic for snapshots so for the time being we need to mock this function out. I wasn't really sure what was the correct way to mock it, I just returned an unchanging string. (LG Ticket is [PD-2179](https://jira.mongodb.org/browse/PD-2179), seems like it will be resolved soon.)

Configuration issues:
* For some reason, storyshots does not like it when I have any `require` statements at the top of the `.storybook/main.js` file. However, it's fine if I have any `require` statements embedded in the Storybook config. This is not great because it goes against the ESLint style guide [rules](https://eslint.org/docs/latest/rules/global-require) but I could not find a way to get both Storybook and storyshots working without doing this. 😔

### Testing 
* Existing tests should pass


src/storyshots.test.ts
```diff
@@ -2,16 +2,26 @@ import initStoryshots, {
...
-      snapshotFileName: snapshotFileName.replace("src", "."),
-    });
-  },
+// Recent LG components introduced a breaking change for storyshots because of the introduction of
+// createUniqueClassName which generates a unique classname at run time. We need to mock out the function
+// so that our snapshots don't break. (https://jira.mongodb.org/browse/PD-2179)
+jest.mock("@leafygreen-ui/lib", () => ({
+  ...jest.requireActual("@leafygreen-ui/lib"),
+  createUniqueClassName: jest.fn().mockReturnValue("static-lg-classname"),
+}));
```
sophstad, line 11: 🔥

package.json
```diff
@@ -24,12 +24,25 @@
...
+    "@emotion/react": "11.9.3",
+    "@emotion/styled": "11.9.3",
+    "@leafygreen-ui/button": "16.0.0",
+    "@leafygreen-ui/emotion": "^4.0.0",
+    "@leafygreen-ui/leafygreen-provider": "2.3.2",
+    "@leafygreen-ui/palette": "3.4.0",
+    "@leafygreen-ui/select": "7.0.0",
+    "@leafygreen-ui/text-input": "10.1.0",
+    "@leafygreen-ui/toggle": "8.0.2",
+    "@leafygreen-ui/typography": "13.1.1",
```
sophstad, line 36: Does this feel like the right PR in which to add fonts? Feel free to say no bc Friday EOD + your vacation next week 😁 If not it looks like we should open a ticket!

src/App.tsx
```diff
@@ -18,6 +19,7 @@ const App = () => {
           Edit <code>src/App.tsx</code> and save to test HMR
         </p>
       </div>
```
sophstad, line 21: IMO we could get rid of all of the above + its unit tests now but your call!

package.json
minnakt: I'll open a ticket! 🙏

sophstad (approved):

PR #9, opened 2022-08-22
EVG-17447 Add Apache license and init cypress dashboard

khelif96:
EVG-17447

### Description 
Add Apache license

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -12,4 +12,5 @@ export default defineConfig({
   },
   viewportWidth: 1920,
   viewportHeight: 1080,
+  videoCompression: false,
 });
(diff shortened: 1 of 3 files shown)


cypress.config.ts
```diff
@@ -11,4 +12,5 @@ export default defineConfig({
   },
   viewportWidth: 1920,
   viewportHeight: 1080,
+  videoCompression: false,
```
sophstad, line 15: I noticed we don't use this setting on Spruce, did you notice the videos were looking bad on the dashboard without it?

sophstad (approved):
not a lawyer but LGTM 😄

cypress.config.ts
khelif96: Yes they looked pretty blurry when i was initially testing.

PR #10, opened 2022-08-23
EVG-17523 Set up route infrastructure

khelif96:
EVG-17523

### Description 
Set up initial dummy routes for parsley. 

### Screenshots
image


src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,18 @@
+import Styled from "@emotion/styled";
+import { palette } from "@leafygreen-ui/palette";
+import { size } from "constants/tokens";
+
+const { gray, white } = palette;
+
+const NavBar: React.FC = () => <Container>OUR TRUSTY NAVBAR</Container>;
+
+const Container = Styled.div`
```
sophstad, line 9: I know this is just a fake navbar, but why not use semantic html 😁 
```suggestion
const Container = Styled.nav`
```

sophstad (approved):

PR #11, opened 2022-08-24
EVG-17470 Set up environmental variables for parsley

khelif96:
EVG-17470

### Description 
Set up environmental variables for the parsley repo.
`.cmdrc.json` can be found on the evergreen-ui channel


package.json
```diff
@@ -5,11 +5,15 @@
   "type": "commonjs",
   "scripts": {
     "build": "tsc && vite build",
-    "build:local": "yarn build",
+    "build:local": "env-cmd -e local -r env/.cmdrc.json yarn build",
```
sophstad, line 8: [nit] Not a requirement, but we could switch to using env-cmd's default config filepath `.env-cmdrc(|.js|.json|)` file if we wanted to avoid repeating the path 6 times

sophstad (approved):

PR #13, opened 2022-08-25
EVG-17511 Create react context that stores log lines and wraps app

khelif96:
EVG-17511

### Description 
Creates a simple context that exposes a hook to ingest and clear log lines
### Screenshots
No visible changes
### Testing 
Wrote unit tests

### DIFF
--- a/src/context/LogContext/index.tsx
+++ b/src/context/LogContext/index.tsx
@@ -36,11 +36,11 @@ const LogContextProvider: React.FC<{ children: React.ReactNode }> = ({
   const memoizedContext = useMemo(
     () => ({
       logLines: state.logs,
-      lineCount: state.lineCount,
+      lineCount: state.logs.length,
       clearLines,
       ingestLines,
     }),
-    [state.logs, state.lineCount, ingestLines, clearLines]
+    [state.logs, ingestLines, clearLines]
   );
 
   return (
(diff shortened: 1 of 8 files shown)


src/context/LogContext/state.ts
```diff
@@ -0,0 +1,42 @@
...
+  lineCount: 0,
+};
+
+const reducer = (state: LogState, action: Action) => {
+  switch (action.type) {
+    case "INGEST_LOGS":
+      return {
+        ...state,
+        logs: action.logs,
+        lineCount: action.logs.length,
```
sophstad, line 21: Do you think `lineCount` is necessary to manage via state? I would argue that because it is directly derived from `logs`, it's more of an intermediate value and would be more straightforward to export a function from the hook that is a log length getter than to manage separately.

khelif96: Good point done!

src/context/LogContext/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  );
+
+  const clearLines = useCallback(() => {
+    dispatch({ type: "CLEAR_LOGS" });
+  }, [dispatch]);
+
+  const memoizedContext = useMemo(
+    () => ({
+      logLines: state.logs,
+      lineCount: state.lineCount,
```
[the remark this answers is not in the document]
khelif96: In the future there may be parts of the state that are internal to the context that we wouldn't want to expose so it seemed to make sense to explicitly return values instead of returning everything.

sophstad (approved):

PR #14, opened 2022-08-25
EVG-17564 Add Mongodb fonts 

khelif96:
EVG-17564

### Description 
Adds MongoDB branding fonts

### Screenshots
image


src/components/styles/fonts.ts
```diff
@@ -0,0 +1,112 @@
...
+    src: url("/static/fonts/EuclidCircularA/EuclidCircularA-MediumItalic-WebXL.woff2")
+        format("woff2"),
+      url("/static/fonts/EuclidCircularA/EuclidCircularA-MediumItalic-WebXL.woff")
+        format("woff");
+  }
+
+  /* Normal */
+  @font-face {
+    font-family: "Euclid Circular A";
+    font-weight: 400, normal;
```
sophstad, line 55: 400 [corresponds to normal](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-weight#common_weight_name_mapping) so this is superfluous (also not positive it's valid CSS)
```suggestion
    font-weight: 400;
```

src/components/styles/fonts.ts
```diff
@@ -0,0 +1,112 @@
...
+    src: url("/static/fonts/EuclidCircularA/EuclidCircularA-Regular-WebXL.woff2")
+        format("woff2"),
+      url("/static/fonts/EuclidCircularA/EuclidCircularA-Regular-WebXL.woff")
+        format("woff");
+  }
+
+  /* Italic */
+  @font-face {
+    font-family: "Euclid Circular A";
+    font-weight: 400, normal;
```
sophstad, line 66: Same here
```suggestion
    font-weight: 400;
```

src/components/styles/fonts.ts
```diff
@@ -0,0 +1,112 @@
+import { css } from "@emotion/react";
```
sophstad, line 1: Perhaps we should omit the Euclid Light and Bold font files since they're not registered via `@font-face` (which is fine — LeafyGreen [doesn't include them](https://github.com/mongodb/leafygreen-ui/blob/c26d1bec13a4fe18612c4a0e36c835cd83e454ef/website/styles/globals.tsx) either)

sophstad (changes requested):

src/components/styles/fonts.ts
khelif96: Thanks just did some copypasta from the mms repo :) I should have checked

sophstad (approved):

PR #17, opened 2022-08-26
EVG-17444 Add query param util functions

khelif96:
EVG-17444

### Description 
* Add util functions for manipulating query params
* Create fully typed hooks that allow us to interact with query params
* Wire up SearchBar with these query param functions.


src/constants/queryParams.ts
```diff
@@ -0,0 +1,8 @@
+enum queryParams {
+  search = "search",
+  highlights = "highlights",
+  bookmarks = "bookmarks",
+  filters = "filters",
+}
+
+export { queryParams };
```
sophstad, line 8: Enums are typically capitalized I believe
```suggestion
enum QueryParams {
  Search = "search",
  Highlights = "highlights",
  Bookmarks = "bookmarks",
  Filters = "filters",
}

export { QueryParams };
```

src/hooks/useQueryParam/index.ts
```diff
@@ -0,0 +1,54 @@
...
+      navigate(`?${stringifiedQuery}`, { replace: true });
+    },
+    [navigate]
+  );
+
+  return [parseQueryString(searchParams.toString()), setQueryString] as const;
+};
+
+/** `useQueryParam` allows you to interact with a query param in the same way you would use a useState hook.  */
+const useQueryParam = <T>(
```
sophstad, line 21: Were there any other particular use cases/types you had in mind when using a generic here?

src/hooks/useQueryParam/index.ts
```diff
@@ -0,0 +1,54 @@
...
+          searchParams[param],
+          Array.isArray(def)
+        ) as unknown as T)
+      : def,
+    setQueryParam,
+  ] as const;
+};
+
+/** `conditionalToArray` takes in a value and transforms it into an array if it is not one and should be */
+const conditionalToArray = <T>(value: T, shouldBeArray: boolean) => {
```
sophstad, line 48: Is `shouldBeArray` superfluous? It seems like in the usage above you make a call to `Array.isArray()` twice, and I feel like the naming/design of this function overall means that we are assuming that an array will be returned. Would this get the job done?

```js
const conditionalToArray = <T>(value: T) => Array.isArray(value) ? value : [value];
```

src/hooks/useQueryParam/index.ts
khelif96: It just allows us to not have to type the type and get strong type safety.

It works in the same way as useState()

if you pass in a string it will ensure only strings are used.
If you pass in an array it will ensure only arrays are used.
number etc.

```ts
// str is of type string, setStr accepts only a string
const [str, setStr] = useQueryParam(queryParam, "something");
// num is of type number, setNum accepts only a number
const [num, setNum] = useQueryParam(queryParam, 1);
// it even allows you to assert a type if it can't be inferred
// For example if you initialize a null or empty array you can tell it that the array should be a number array
const [someArray, setSomeArray] = useQueryParam<number[]>(queryParam, [])
```

src/hooks/useQueryParam/index.ts
khelif96: not necessarily if our default type is an array we are able to determine that this type should always be an array and if we should return an array. 

The reason this is necessary is that `query-param` does not know if a value is an array unless it has 2 or more values. 
so 

`?someArray=1`
This would always return the value `1` instead of `[1]`

`?someArray=1,2`
This would return `[1,2]` 

By adding the shouldBeArray flag we can programmatically determine if our resulting value should be an array because the default value was an array. We can't only rely on just the current value.

sophstad (approved):

khelif96:
Checks passed but weren't reported. https://spruce.mongodb.com/version/630fb39f1e2d173d5c51b9bf/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #28, opened 2022-09-01
EVG-17514 Add LogWindow using react-virtualized

khelif96:
EVG-17514

### Description 
Adds a window that will render logs
* Created placeholder components for sidebar and subheader
* Created placeholder row components 
* Add support for line wrapping. (This should be controlled by some other state mechanism but for now using query params)
* Double clicking on a row selects it and adds it to the url
* Selected elements will be scrolled to automatically 
* Added cypress utilities for determining if an element is in view.
  * This is necessary since cypress assertions can't tell if an item is partially in view. Since we have a horizontal scroll aspect we need a method to determine if an element is partially in the window or fully in view.
* Added placeholder logs and emulated loading with a setTimeout

### Screenshots
image

https://user-images.githubusercontent.com/4605522/187964947-03e21c2d-b7b9-4c81-8d90-ecaf6b72e780.mov

### Testing 
Wrote unit tests for each component
and e2e tests


src/components/LogPane/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type LogPaneProps = Omit<
+  ListProps,
+  "height" | "width" | "itemData" | "rowHeight"
+> & {
+  logLines: string[];
+  wrap: boolean;
+};
+
+const LogPane: React.FC<LogPaneProps> = ({
```
minnakt, line 12: Very mysterious but I think the last log line isn't getting rendered properly 🤔


https://user-images.githubusercontent.com/47064971/188020129-697c7aa0-2dbc-4fc0-8d12-6d599c097900.mov

I think there should be a line saying `[j0:sec1] Starting mongod on port 20002...` at the end

minnakt (changes requested):
looking good so far 😃

src/components/LogRow/ResmokeRow/ResmokeRow.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { renderWithRouterMatch, screen, userEvent, waitFor } from "test_utils";
+import { ResmokeRow } from ".";
+
+const wrapper = (logs: string[]) =>
+  // eslint-disable-next-line func-names, react/display-name
+  function ({ children }: { children: React.ReactNode }) {
+    return (
+      <LogContextProvider initialLogLines={logs}>{children}</LogContextProvider>
+    );
+  };
```
minnakt, line 12: Alternatively maybe like this?

```suggestion
const wrapper = (logs: string[]) => {
  const provider = ({ children }: { children: React.ReactNode }) => (
    <LogContextProvider initialLogLines={logs}>{children}</LogContextProvider>
  );
  return provider;
}
```

src/components/SubHeader/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const StyledIcon = styled(Icon)`
+  padding-left: ${size.l};
+  padding-right: ${size.l};
+`;
+
+const Container = styled.div`
+  background-color: ${gray.light3};
+  width: calc(100vw - ${size.xl});
+  box-shadow: 0 4px 4px rgba(0, 0, 0, 0.05);
```
minnakt, line 25: nit. use tokens?

src/components/SideBar/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+);
+
+const StyledButton = styled(Button)`
+  width: 55px;
+`;
+const Container = styled.div`
+  background-color: ${gray.light3};
+  width: ${size.xl};
+  max-height: calc(100vh - ${navbarHeight});
+  box-shadow: 0 4px 4px rgb(0 0 0 / 25%);
```
minnakt, line 23: nit. use tokens?

minnakt (changes requested):
There are a few places where it seems like React is unnecessarily imported (`import React from "react"`), would you mind going through and removing them?

src/components/LogRow/ResmokeRow/ResmokeRow.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+  (Story) => (
+    <MemoryRouter initialEntries={["/"]}>
+      <Story />
+    </MemoryRouter>
+  ),
+];
+
+const MultiLineTemplate: ComponentStory<typeof ResmokeRow> = (args) => (
+  <>
+    {logLines.map((value, index) => (
```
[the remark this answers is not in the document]
khelif96: I'm surprised eslint didn't catch this...

src/components/LogRow/AnsiiRow/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  return (
+    <CellMeasurer key={key} cache={cache} parent={parent} rowIndex={index}>
+      <AnsiiRow {...props} />
+    </CellMeasurer>
+  );
+};
+
+const cache = new CellMeasurerCache({
+  fixedWidth: true,
+  defaultHeight: 16,
```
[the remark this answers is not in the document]
khelif96: This is the height it uses when an element hasn't rendered yet afaict. It assumes the element is this height before it calculates the real height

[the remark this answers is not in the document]
khelif96: It can be any value and it would just affect the size of the scroll bar on the side.

khelif96:
The best i can gather for why the tests are failing is that ubuntu is rendering the text differently

cypress/support/commands.ts
```diff
@@ -27,3 +29,53 @@ Cypress.Commands.add(
...
+    const right = window.width();
+    const rect = subject[0].getBoundingClientRect();
+
+    // All corners of the element must be in the viewport
+    expect(rect.top).not.to.be.greaterThan(bottom);
+    expect(rect.bottom).not.to.be.greaterThan(bottom);
+    expect(rect.left).not.to.be.greaterThan(right);
+    expect(rect.right).not.to.be.greaterThan(right);
+
+    return subject;
```
minnakt, line 49: How about something like this?

```
    const rect = subject[0].getBoundingClientRect();
    const width = Cypress.config().viewportWidth;
    const height = Cypress.config().viewportHeight;

    // All corners of the element must be in the viewport
    expect(rect.top).to.be.within(0, height);
    expect(rect.right).to.be.within(0, width);
    expect(rect.bottom).to.be.within(0, height);
    expect(rect.left).to.be.within(0, width);

    return subject;
```

The tests passed when I tried it, though I'm not entirely sure what the difference is. ([link](https://spruce.mongodb.com/task/parsley_ubuntu1604_large_e2e_test_patch_84f5549e43a931d909770ab50065720bb8be2c8c_631763291e2d17647d8a7d1f_22_09_06_15_11_44/logs?execution=0))

src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { ResmokeRow } from "../ResmokeRow";
+
+interface RowProps {
+  listRowProps: ListRowProps;
+  data: {
+    getLine: (index: number) => string | undefined;
+    wrap: boolean;
+    processedLines: (number | number[])[];
+  };
+}
```
minnakt, line 18: Kind of wondering if this belongs better in `BaseRow/index.tsx` or the `types.ts` file in this folder

minnakt (changes requested):
It looks good, though I'd want to look at it one more time before approving

cypress/support/commands.ts
khelif96: I think this might have to do with how chrome is rendering the text instead of an issue with the test itself

src/components/LogRow/RowRenderer/index.tsx
khelif96: Good call done

cypress/support/commands.ts
khelif96: So the bug was actually caused by the scrollbar of the page taking up space on linux. Macos treats the scrollbar as an overlay while linux treats it as an element with a set width.

src/components/LogRow/BaseRow/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  margin-top: 0;
+  margin-bottom: 0;
+  /* margin-left: ${size.xs}; */
+  margin-right: ${size.xs};
+  font-size: ${fontSize.m};
+  :before {
+    content: "${(props) => props.index}";
+    margin-left: ${size.s};
+    margin-right: ${size.xl};
+  }
```
minnakt, line 66: nice 🐠

minnakt (approved):
reel good work here 🎣😎

khelif96:
Tests passed :) https://spruce.mongodb.com/version/6317bd053627e067aad46eee/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #30, opened 2022-09-06
EVG-17633 Attach dist

khelif96:
EVG-17633

### Description 
Just attaches the production dist folder to the evergreen task


khelif96:
The s3 bucket the dist is uploaded to isn't really configured to serve SPA's so it isn't worth it to serve the app in this way.

PR #31, opened 2022-09-08
EVG-15938 Add File uploader

khelif96:
EVG-15938

### Description 
Users can now upload log files using the file picker or via drag and drop.
Added a confirmation dialog for navigating to the file uploader since it will reset the page state.  We don't have designs for this but it seemed useful.

### Screenshots

https://user-images.githubusercontent.com/4605522/189225477-11ea6a03-1fce-4dff-94ea-713812a2a901.mov

### Testing 
Unit and e2e tests


src/pages/LogDrop/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+        value={logType}
+      >
+        <Option key="resmoke" value={LogTypes.RESMOKE_LOGS}>
+          Resmoke
+        </Option>
+        <Option key="standard" value={LogTypes.EVERGREEN_TASK_LOGS}>
+          Raw
+        </Option>
+      </StyledSelect>
+      <Button data-cy="process-log-button" onClick={onParse} variant="danger">
```
sophstad, line 65: This doesn't feel like the right use case for a danger variant to me 🤔

src/pages/LogDrop/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+        Process Log
+      </Button>
```
sophstad, line 67: It may be helpful to also include a "Cancel" button here in case the user accidentally uploaded the wrong file. They could be taken back to the upload page.
image

src/components/SubHeader/index.tsx
```diff
@@ -7,10 +7,17 @@ import { size, subheaderHeight } from "constants/tokens";
...
-const SubHeader = () => (
+interface SubHeaderProps {
+  isUploadedLog: boolean;
+}
+const SubHeader: React.FC<SubHeaderProps> = ({ isUploadedLog }) => (
   <Container>
     <StyledIcon glyph="EvergreenLogo" />
-    <StyledLink href="/test">Task Page</StyledLink>
+    {isUploadedLog ? (
+      <span>Thanks for uploading</span>
```
sophstad, line 17: This seems somewhat irrelevant to pin to the page as a non-dismissible banner. Maybe showing the filename of the uploaded log would be helpful here?

src/pages/LogDrop/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+      >
+        <Option key="resmoke" value={LogTypes.RESMOKE_LOGS}>
+          Resmoke
+        </Option>
+        <Option key="standard" value={LogTypes.EVERGREEN_TASK_LOGS}>
+          Raw
+        </Option>
+      </StyledSelect>
+      <Button data-cy="process-log-button" onClick={onParse} variant="danger">
+        Process Log
```
sophstad, line 66: This button is currently clickable when no option is selected. It should be disabled until a log type is chosen.

src/pages/LogDrop/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+const { red } = palette;
+interface LogDropProps {
+  onChangeLogType: (logType: LogTypes) => void;
+}
+
+const LogDrop: React.FC<LogDropProps> = ({ onChangeLogType }) => {
+  const { ingestLines } = useLogContext();
+  const [hasDroppedLog, setHasDroppedLog] = useState(false);
+  const [logType, setLogType] = useState(LogTypes.RESMOKE_LOGS);
```
sophstad, line 21: Since we're not doing any type of file detection, it feels more intuitive to me to default to no selection here and require users to choose the type of log they've uploaded.

src/pages/LogDrop/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  return hasDroppedLog ? (
+    <Container>
+      <StyledSelect
+        label="How would you like to parse this log?"
+        onChange={(value) => setLogType(value as LogTypes)}
+        value={logType}
+      >
+        <Option key="resmoke" value={LogTypes.RESMOKE_LOGS}>
```
sophstad, line 58: [nit] I don't think you need a `key` here, seems like LeafyGreen handles that.

src/pages/LogDrop/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+      {isDragActive ? (
+        <Body>Release to upload</Body>
+      ) : (
+        <Dropzone>
+          <Body weight="medium">
+            Drag and Drop a log file to view in Lobster
+          </Body>
+          <Body weight="medium">or</Body>
+          <Button leftGlyph={<Icon glyph="Upload" />} onClick={open}>
+            Select From Files
```
sophstad, line 81: Just noticed this in the [Button design guidelines](https://www.mongodb.design/component/button/guidelines/):

> ‎‎‎‎‏‏‎Keep button content sentence case. Only capitalize the first letter of the CTA and the first letter of a proper noun, like Atlas, Data Lake, Ops Manager, etc.

```suggestion
            Select from files
```

src/pages/LogDrop/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+  width: 60%;
+  height: 250px;
+  justify-content: center;
+  align-items: center;
+  border: ${size.xxs} dashed ${red.base};
+  border-radius: ${size.s};
```
sophstad, line 97: [nit] We should apply a background to the navbar to avoid this kind of weirdness
image

sophstad (changes requested):

src/pages/LogDrop/index.tsx
khelif96: The red fit in with the rest of the ui we don't actually have a design for this so i took some creative liberty :) But i can make it green.

src/components/SubHeader/index.tsx
khelif96: Good idea i put it on the agenda for next design sync but filename sgtm

src/pages/LogDrop/index.tsx
khelif96: Good catch  i recall react used to complain possibly in an older version of select.

src/components/NavBar/index.tsx
```diff
@@ -18,6 +20,7 @@ const NavBar: React.FC = () => {
     QueryParams.Filters,
     []
   );
+  const { hasLogs, clearLines } = useLogContext();
```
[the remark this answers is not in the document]
khelif96: It was easier to unit test :)

src/pages/LogDrop/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+const ButtonContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+  margin-top: ${size.xs};
+  button {
+    margin: 0 ${size.xs};
+  }
+`;
+const Dropzone = styled.div`
```
minnakt, line 118: nit. some more vertical spacing would be nice! maybe 

```
> * {
    margin-bottom: ${size.xs};
}
```

Screen Shot 2022-09-09 at 11 59 53 AM

src/components/SubHeader/index.tsx
```diff
@@ -3,16 +3,28 @@ import { palette } from "@leafygreen-ui/palette";
...
-);
+interface SubHeaderProps {
+  isUploadedLog: boolean;
+}
+const SubHeader: React.FC<SubHeaderProps> = ({ isUploadedLog }) => {
+  const { fileName } = useLogContext();
+
+  return (
+    <Container>
+      <StyledIcon glyph="EvergreenLogo" />
```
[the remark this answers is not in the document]
khelif96: Great catch it looks like the reset styles are affecting it somehow.

src/pages/LogDrop.tsx
```diff
@@ -0,0 +1,32 @@
+import { useState } from "react";
+import style from "@emotion/styled";
+import { PageLayout } from "components/styles";
+import { LogTypes } from "constants/enums";
+import { useLogContext } from "context/LogContext";
+import FileDropper from "./LogDrop/index";
```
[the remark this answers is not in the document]
khelif96: Good call out! i forgot to clean up some files i've moved some things around.

src/pages/LogDrop/index.tsx
khelif96: The navbar does have a background color. I think the container being to large was causing some weird overlapping. Never the less i made the height a percentage and gave it a max value so it doesn't become huge.  This should hopefully be a nice compromise and it avoids this in most scenarios.

src/components/SubHeader/index.tsx
```diff
@@ -3,26 +3,34 @@ import { palette } from "@leafygreen-ui/palette";
...
-);
+const SubHeader: React.FC<SubHeaderProps> = ({ isUploadedLog }) => {
+  const { fileName } = useLogContext();
 
-const StyledIcon = styled(Icon)`
-  padding-left: ${size.l};
+  return (
+    <Container>
+      {isUploadedLog ? (
+        <span>File name: ./{fileName}</span>
```
sophstad, line 20: [nit] I would probably omit the `./` here, feel like it implies some type of filesystem/hierarchy where there is none. And developers can be finicky 😁

package.json
```diff
@@ -36,6 +36,7 @@
     "@emotion/styled": "11.9.3",
     "@leafygreen-ui/button": "16.0.0",
     "@leafygreen-ui/card": "7.0.0",
+    "@leafygreen-ui/confirmation-modal": "^3.0.6",
```
sophstad, line 39: Should we pin this dep?

sophstad (approved):
🎉

src/components/LogWindow/index.tsx
```diff
@@ -26,10 +27,10 @@ const LogWindow: React.FC<LogWindowProps> = ({ logType }) => {
   );
 
   return (
-    <Container>
-      {logLines.length !== 0 && <SideBar maxLineNumber={logLines.length - 1} />}
+    <Container data-cy="log-window">
+      {hasLogs && <SideBar maxLineNumber={logLines.length - 1} />}
```
minnakt, line 31: thank u king

minnakt (approved):

PR #32, opened 2022-09-08
EVG-17512: Create Details Overlay card

minnakt:
[EVG-17512](https://jira.mongodb.org/browse/EVG-17512)

### Description 
This PR adds the `DetailsOverlay` card.

### Notes
I needed to add labels to the toggles because it didn't quite make sense for Filter Logic, which is one of `"and"` or `"or"`, to correspond to an on/off toggle.

The visibility and toggles for the filters aren't hooked up anywhere. I wasn't sure what was the best way to store this information but it could be done as part of [EVG-17537](https://jira.mongodb.org/browse/EVG-17537).

### Screenshots
Screen Shot 2022-09-12 at 4 41 09 PM

### Testing 
- Added `DetailsOverlay.test.tsx`
- Edited the Cypress tests for wrapping to not be manual, instead toggling the setting in the `DetailsOverlay` card


package.json
```diff
@@ -107,7 +110,7 @@
     "replace-in-file": "^6.3.5",
     "rollup-plugin-visualizer": "^5.8.0",
     "ts-jest": "28.0.7",
-    "typescript": "^4.6.4",
+    "typescript": "4.8.2",
```
khelif96, line 113: Sneaking in a typescript update 👁️ 👁️

src/components/DetailsOverlay/SearchRangeInput/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+      setLowerQueryParam(parseInt(newVal, 10));
+    }
+  };
+
+  const updateUpperBound = (newVal: string) => {
+    if (newVal === "") {
+      setUpperBound(undefined);
+    } else {
+      setUpperBound(parseInt(newVal, 10));
+    }
```
khelif96, line 31: I think setting this to -1 should remove the upperbound

src/components/DetailsOverlay/SearchRangeInput/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  return (
+    <FilterRow>
+      <StyledSubtitle> Range </StyledSubtitle>
+      <RangeContainer>
+        <RangeInput
+          aria-labelledby="Range Lower Bound"
```
khelif96, line 39: I think this should be aria-label.
 https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

src/components/DetailsOverlay/SearchRangeInput/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+          data-cy="range-lower-bound"
+          min={0}
+          onChange={(e) => updateLowerBound(e.target.value)}
+          placeholder="0"
+          sizeVariant="small"
+          type="number"
+          value={(lowerBound ?? "").toString()}
+        />
+        <RangeInput
+          aria-labelledby="Range Upper Bound"
```
khelif96, line 49: Same comment

src/components/DetailsOverlay/Toggle/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+        aria-labelledby="base-toggle"
+        checked={value}
+        data-cy={dataCy}
+        onChange={onChange}
+        size="small"
+      />
+      <StyledDisclaimer>{rightLabel}</StyledDisclaimer>
+    </ToggleWithLabel>
+  </FilterRow>
+);
```
khelif96, line 46: Maybe this should be its own component?  Some Storybook stories and tests would be nice. :)

src/components/DetailsOverlay/Toggle/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  return (
+    <BaseToggle
+      data-cy="format-v2-toggle"
+      label="Format LogV2"
+      onChange={onChange}
+      value={logV2}
+    />
+  );
+};
```
khelif96, line 90: I spoke to @sanjogDB  lets omit this feature i think we will only support the resmoke v2 format since the legacy format isn't used anymore.

src/components/DetailsOverlay/Toggle/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+export const PrettyPrintToggle: React.FC = () => {
+  const [prettyPrint, setPrettyPrint] = useState(
+    Cookie.get(PRETTY_PRINT_BOOKMARKS) === "true"
+  );
+  const onChange = (checked: boolean) => {
+    setPrettyPrint(checked);
+    Cookie.set(PRETTY_PRINT_BOOKMARKS, checked.toString(), { expires: 365 });
+  };
```
khelif96, line 141: How do we signal to the rest of the app that this value changed? We might need to hook this into context. But thats a problem for https://jira.mongodb.org/browse/EVG-17682

src/components/DetailsOverlay/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) => {
+  const [filters, setFilters] = useQueryParam<string[]>(
+    QueryParams.Filters,
+    []
+  );
+
+  const deleteFilter = (filterName: string) => {
+    const newFilters = filters.filter((f) => f !== filterName);
+    setFilters(newFilters);
+  };
```
khelif96, line 32: Deleting doesn't seem to work


https://user-images.githubusercontent.com/4605522/190255398-94ed05cc-646a-472d-a721-7116597dd2e9.mov

src/components/DetailsOverlay/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+interface DetailsOverlayProps {
+  ["data-cy"]?: string;
+}
+
+const DetailsOverlay: React.FC<DetailsOverlayProps> = ({
+  "data-cy": dataCy,
+}) => {
+  const [filters, setFilters] = useQueryParam<string[]>(
+    QueryParams.Filters,
+    []
```
khelif96, line 26: These filters are deduped like the below video.

src/components/DetailsOverlay/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <ExpandableRowsToggle />
+          <PrettyPrintToggle />
+          <ButtonRow />
+        </FilterContainer>
+      </FilterOptions>
+
+      {filters.length !== 0 && <Divider />}
+      {filters.map((filter) => (
+        <Filter key={filter} deleteFilter={deleteFilter} filterText={filter} />
+      ))}
```
khelif96, line 54: Allot of filters overlap the page and cause us to overflow.
Limiting the height of the container may help/ introducing some scroll.
image

src/components/Filter/Toggle/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  leftText: string;
+  rightText: string;
+}
+
+const Toggle: React.FC<ToggleProps> = ({
+  activeButton = "left",
+  "data-cy": dataCy,
+  leftText,
+  rightText,
+}) => {
```
khelif96, line 23: I think making this a controlled component would be the easiest to use.

So adding a `value` and `onChange` prop. And using those to manage its state.

src/components/DetailsOverlay/styles.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const FilterRow = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin-bottom: ${size.m};
+`;
+
+// @ts-ignore-error
+export const StyledSubtitle = styled(Subtitle as div)`
```
khelif96, line 13: idk how true this alias is but we should totally steal these from mms https://github.com/10gen/mms/blob/master/client/js/common/types/emotion.ts

khelif96 (changes requested):
This looks great so far!

src/components/DetailsOverlay/ButtonRow/index.tsx
```diff
@@ -0,0 +1,13 @@
+import Button from "@leafygreen-ui/button";
+import Icon from "components/Icon";
+import { FilterRow } from "../styles";
+
+const ButtonRow: React.FC = () => (
+  <FilterRow>
+    <Button leftGlyph={<Icon glyph="Copy" />}> JIRA </Button>
+    <Button leftGlyph={<Icon glyph="Export" />}> RAW </Button>
+    <Button leftGlyph={<Icon glyph="Export" />}> HTML </Button>
```
sophstad, line 9:
```suggestion
    <Button leftGlyph={<Icon glyph="Copy" />}>JIRA</Button>
    <Button leftGlyph={<Icon glyph="Export" />}>RAW</Button>
    <Button leftGlyph={<Icon glyph="Export" />}>HTML</Button>
```

src/components/NavBar/index.tsx
```diff
@@ -48,7 +49,9 @@ const NavBar: React.FC = () => {
       </FlexContainer>
       <ButtonContainer>
         <StyledButton buttonText="Filters">SomeContent</StyledButton>
-        <StyledButton buttonText="Details">SomeContent</StyledButton>
+        <StyledButton buttonText="Details" data-cy="details-button">
```
sophstad, line 52: It'd be awesome to have a badge/circle/icon of some sort with the number of filters applied when the dropdown is closed. Maybe worth a new ticket or design discussion though?

src/components/DetailsOverlay/Toggle/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+    <BaseToggle
+      data-cy="format-v2-toggle"
+      label="Format LogV2"
+      onChange={onChange}
+      value={logV2}
+    />
+  );
+};
+
+export const FilterLogicToggle: React.FC = () => {
```
sophstad, line 92: This almost feels like the right use case for a radio button? It would allow us to remove the labels from the other options. Your call though!

image

src/components/DetailsOverlay/SearchRangeInput/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    if (newVal === "") {
+      setUpperBound(undefined);
+    } else {
+      setUpperBound(parseInt(newVal, 10));
+    }
+  };
+
+  return (
+    <FilterRow>
+      <StyledSubtitle> Range </StyledSubtitle>
```
sophstad, line 36:
```suggestion
      <StyledSubtitle> Range </StyledSubtitle>
```

src/components/DetailsOverlay/Toggle/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+const BaseToggle: React.FC<BaseToggleProps> = ({
+  "data-cy": dataCy,
+  rightLabel = "ON",
+  leftLabel = "OFF",
+  label,
+  value,
+  onChange,
+}) => (
+  <FilterRow>
+    <StyledSubtitle> {label} </StyledSubtitle>
```
sophstad, line 33:
```suggestion
    <StyledSubtitle>{label}</StyledSubtitle>
```

src/components/Filter/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <StyledIcon
+          glyph="X"
+          onClick={() => deleteFilter(filterText)}
+          size={Size.XLarge}
+        />
+        <StyledIcon
+          glyph={isVisible ? "Visibility" : "ClosedEye"}
+          onClick={() => toggleVisible()}
+          size={Size.XLarge}
+        />
```
sophstad, line 38: If not too much trouble, it could be nice to use these icons with an LG `IconButton` to leverage the hover states they've implemented

src/components/DetailsOverlay/SearchRangeInput/index.tsx
minnakt: The Leafygreen input will throw an error if it's `aria-label` with `type="number"`, it only accepts `aria-labelledby`

src/components/DetailsOverlay/SearchRangeInput/index.tsx
minnakt: Hmm when I do that it adds `upper=-1` to the URL. It could be best to just remove it from the URL if it's not a valid upper bound

minnakt:
The filters should be moved to a side panel in [EVG-17908](https://jira.mongodb.org/browse/EVG-17908)

src/components/DetailsOverlay/index.tsx
khelif96: Just wanted to surface this again @minnakt  i think this is still a problem

src/components/DetailsOverlay/Toggle/index.tsx
khelif96: I personally am a fan of the toggles over the radio buttons they seem more intuitive.

src/components/DetailsOverlay/SearchRangeInput/index.tsx
khelif96: I think this just needs a conditional something like...
```ts
// replace this with a better name :) 
const newValNumber = parseInt(newVal,10);
if (newValNumber === -1){
   setUpperBound(undefined);
} else{
   setUpperBound(newValNumber);
}
```

khelif96, line 31: This also doesn't prevent against upperBound being less than lowerBound.
image

src/components/FiltersOverlay/Filter/Toggle/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface ToggleProps {
+  value: boolean;
+  onChange: (newValue: boolean) => void;
+  ["data-cy"]?: string;
+  leftText: string;
+  rightText: string;
+}
+
+const Toggle: React.FC<ToggleProps> = ({
```
khelif96, line 13: Have you explored? Not suggesting it just asking what you think?
https://www.mongodb.design/component/segmented-control/documentation/

khelif96 (changes requested):

src/components/DetailsOverlay/SearchRangeInput/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  );
+  const [upperBound, setUpperBound] = useQueryParam<number | undefined>(
+    QueryParams.UpperRange,
+    undefined
+  );
+
+  const hasError =
+    lowerBound !== undefined && upperBound !== undefined
+      ? lowerBound > upperBound
+      : false;
```
khelif96, line 24: Although we now display an error this doesn't actually prevent the user from using these values. They are still applied to the search and are used. (Although i guess we check this during the search phase) what do you think?

khelif96: Great point i think you are right search and these associated search params don't really make sense as a shareable piece of data. Maybe they should live in context instead. I can address it in my upcoming search pr if you would like

khelif96 (approved):

PR #35, opened 2022-09-09
EVG-17515 Load logs from prod and staging

khelif96:
EVG-17515

### Description 
We can now load logs in production and staging.
Visit https://parsley-beta.corp.mongodb.com/evergreen/spruce_ubuntu1604_e2e_test_6a3b2749a20e97bc08f6cf0c953ba791e5cc54c2_1f7cf49f4ce587c74212d8997da171c4_22_09_07_19_05_15/0/task?selectedLine=2174443&wrap=true
### Screenshots
image

### Testing 
Still trying to figure out the testing story for this so stay tuned...


src/hooks/useAxios/useAxios.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  statusText: "OK",
+  data: jsonMessage,
+};
+
+jest.mock("axios");
+
+describe("useAxiosGet", () => {
+  it("gets a good response from the api and updates its state", async () => {
+    // @ts-expect-error
+    axios.get.mockResolvedValue(mockApi);
```
minnakt, line 20: to get rid of the error, let's do

```
const mockedAxiosGet = axios.get as jest.MockedFunction<typeof axios.get>;
mockedAxiosGet.mockResolvedValue(mockApi);
```

src/pages/LoadingPage.tsx
```diff
@@ -1,3 +1,70 @@
...
+import { useParams } from "react-router-dom";
+import { LogTypes } from "constants/enums";
+import {
+  getEvergreenTaskLogURL,
+  getEvergreenTestLogURL,
+  getResmokeLogURL,
+} from "constants/logURLTemplates";
+import { slugs } from "constants/routes";
+import { useLogContext } from "context/LogContext";
+import { useAxiosGet } from "hooks/useAxios";
```
minnakt, line 11: I think we might want to import this `from "hooks"` (unless you want to get rid of the index file in `/hooks`)

src/utils/routes/index.ts
```diff
@@ -0,0 +1,4 @@
+// Remove : from slug
+const slugToVariable = (slug: string) => slug.replace(":", "");
+
+export { slugToVariable };
```
minnakt, line 4: This could potentially be deleted, as I don't see it being used elsewhere in the code

src/pages/LoadingPage.tsx
```diff
@@ -1,3 +1,70 @@
...
+  useEffect(() => {
+    if (data) {
+      ingestLines(data.split("\n"), logType);
+      onLoad();
+    }
+  }, [data, ingestLines, onLoad, logType]);
+  if (error) {
+    console.error(error);
+  }
+  return <div>I am the loading page</div>;
```
minnakt, line 67: (disclaimer: i did not test this)

I wonder if this needs to be in a hook... 🤔 It would be awesome if we could just do 

```suggestion
  if (data) {
    ingestLines(data.split("\n"), logType);
    onLoad();
  }
  if (error) {
    console.error(error);
  }
  return <div>I am the loading page</div>;
```

src/constants/logURLTemplates.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ *
+ * @param buildID - the build ID of the resmoke job
+ * @param testID - the test ID of the resmoke job
+ * @returns `/build/${buildID}/test/${testID}`
+ */
+const getResmokeLogURL = (buildID: string, testID?: string) => {
+  const params = {
+    raw: 1,
+    s3: true, // TODO: Remove this once this becomes the default
```
minnakt, line 27: nit. Maybe if we have a ticket number for this, we can add it :0

minnakt (changes requested):
This review is getting kind of long, so I'm going to review the scripts in the next pass!

I especially like the improvements in Cypress testing! 😇

I do see some weirdness when loading a log with wrap enabled (see below image), but I suspect that it's because of [EVG-17931](https://jira.mongodb.org/browse/EVG-17931). Could you check a resmoke log and confirm that this is the case? 🙏

Screen Shot 2022-09-27 at 1 34 59 PM

khelif96:
> This review is getting kind of long, so I'm going to review the scripts in the next pass!
> 
> I especially like the improvements in Cypress testing! 😇
> 
> I do see some weirdness when loading a log with wrap enabled (see below image), but I suspect that it's because of [EVG-17931](https://jira.mongodb.org/browse/EVG-17931). Could you check a resmoke log and confirm that this is the case? 🙏
> 
> Screen Shot 2022-09-27 at 1 34 59 PM

Good catch i think this should fix it. https://github.com/evergreen-ci/parsley/pull/50

README.md
```diff
@@ -27,15 +27,10 @@ Parsley is capable of fetching logs from both [evergreen](https://github.com/eve
...
-    cd logkeeper
-    
-    # Use this (or similar command) to seed some sample data into a local bucket and use that as storage
-    mkdir -p _bucketdata && cp -r testdata/simple/* _bucketdata
-    ```
-3. Run the following command to inject the necessary env variables and start the local logkeeper server 
-
-    `LK_CORS_ORIGINS=http:\/\/localhost:\\d+ go run main/logkeeper.go --localPath _bucketdata` 
+2. Run `yarn bootstrap-logkeeper` to download some sample resmoke logs from s3. 
+3. Run the command outputted by the preview step to seed the env variables and start the local logkeeper server
```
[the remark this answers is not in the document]
khelif96: previous* Thanks!

src/constants/logURLTemplates.ts
khelif96: I think it was removed between when i started this pr and now. Removed it!

src/pages/LoadingPage.tsx
khelif96: useEffect is meant to handle side effects like network requests and stuff. Without the hook i think error would trigger every render i've moved it into the hook.

khelif96:
Set module patch https://spruce.mongodb.com/version/6334503f57e85a52e060003d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

cypress/integration/ansiiLogs/ansii_logView.ts
```diff
@@ -31,41 +34,44 @@ describe("Basic evergreen log view", () => {
     cy.dataCy("log-row-2").should("be.visible");
     cy.dataCy("log-row-2").should(
       "contain.text",
-      "disableLogicalSessionCacheRefresh"
+      `[2022/03/02 17:01:58.701] Running pre-task commands.`
     );
     cy.dataCy("log-row-2").isContainedInViewport();
```
minnakt, line 39: I actually think you want these to all be `log-row-22` rather than `log-row-2`, based on the corresponding resmoke test. (i.e. in the previous test we check that `log-row-22` is not captured in the viewport due to wrap being disabled, in this test we want to test that `log-row-22` is captured in the viewport due to wrap being enabled)

scripts/bootstrap-logkeeper.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "Uncompressing bucket data..."
+    tar -xzf ./bin/_bucketdata.tar.gz -C ./bin/
+    # Check to see if the uncompress was successful
+    if [ $? -ne 0 ]; then
+        echo "${RED}Failed to uncompress bucket data!${NC}"
+        echo "Cleaning up _bucketdata directory..."
+        rm -rf _bucketdata
+        exit 1
+    fi
+    echo "Uncompressed files."
```
minnakt, line 39: For clarity

```suggestion
    echo "Finished uncompressing files."
```

scripts/bootstrap-logkeeper.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "Cleaning up _bucketdata directory..."
+    rm bin/_bucketdata.tar.gz
+    echo "Done!"
+    echo "${GREEN}Bucket data downloaded successfully!${NC}"
+    
+else
+    echo "Found bin/_bucketdata directory, skipping download..."
+    echo "If you want to download the bucket data again, delete the _bucketdata directory and run this script again."
```
minnakt, line 47: let's get rid of the redundant `again`s

```suggestion
    echo "If you want to download the bucket data again, delete the _bucketdata directory and rerun this script."
```

src/components/LogPane/index.tsx
```diff
@@ -32,10 +32,11 @@ const LogPane: React.FC<LogPaneProps> = ({
         <List
           // If wrap is false, We want the list to scroll horizontally.
```
minnakt, line 33: nit. I think you can remove this comment since we are no longer conditionally applying the `containerStyle`

cypress.config.ts
```diff
@@ -7,6 +7,8 @@ export default defineConfig({
     supportFile: "cypress/support/index.ts",
     specPattern: "cypress/integration/**/*.ts",
     experimentalStudio: true,
+    experimentalSessionAndOrigin: true,
+    testIsolation: "legacy",
```
minnakt, line 11: I feel like you can delete this setting since it seems to be the default.

```suggestion

```

cypress/integration/ansiiLogs/ansii_logView.ts
khelif96: oops great catch!

minnakt (approved):
🔥🔥🔥

PR #36, opened 2022-09-09
EVG-17524 Add ANSII encoded log line support

khelif96:
EVG-17524

### Description 
Support ansii encoded lines and linkify log lines.
Updated the storybook styles to have similar horizontal scroll behavior as the log panes.
### Screenshots
image


src/components/LogRow/AnsiiRow/AnsiiRow.test.tsx
```diff
@@ -78,6 +78,19 @@ describe("ansiiRow", () => {
       expect(history.location.search).toBe("");
     });
   });
+  it("lines should be linkified if they have a url", async () => {
```
minnakt, line 81: nit. I think you don't need `async`

src/components/LogRow/ResmokeRow/ResmokeRow.stories.tsx
```diff
@@ -46,7 +46,9 @@ SingleLine.decorators = [
 ];
 
 const MultipleLineTemplate: ComponentStory<ResmokeRowProps> = (args) => (
-  <>
+  <div
+    style={{ overflow: "scroll visible", position: "relative", height: 500 }}
+  >
     {logLines.map((_, index) => (
```
minnakt, line 52: For some reason, `wrap` is broken on this story

src/components/LogRow/AnsiiRow/index.tsx
```diff
@@ -11,9 +16,20 @@ const AnsiiRow = forwardRef<any, BaseRowProps>((rowProps, ref) => {
...
+  const memoizedLogLine = useMemo(() => {
+    if (!lineContent) return null;
+    const render = linkifyHtml(ansiUp.ansi_to_html(lineContent), {
+      validate: {
+        url: (value: string) => /^(http)s?:\/\//.test(value),
+      },
+    });
+    return parse(render);
+  }, [lineContent]);
+  if (!lineContent) return null;
```
minnakt, line 29: I have two related questions:
1. Does `getLine` need to be typed as returning `string | undefined`? Because if not, I think we wouldn't need to check if `lineContent` is undefined
2. If it does need to be typed that way, do we need the line `if (!lineContent) return null;` twice?

minnakt (changes requested):

src/components/LogRow/AnsiiRow/index.tsx
khelif96: 1. Yes because  in the chance we try to fetch a log line and we don't have its index we should return undefined. As a defensive tactic.
2. So `ansi_to_html` needs a value passed in so we need a guard to ensure we only pass in non null values. We also need a guard on line 29 to not return anything for a row if there isn't any content. That can probably change in the future. We can't move the guard up to after line 19 since that would break the rules of hooks. So it seemed the least hacky to just have 2 guards.

src/components/LogRow/AnsiiRow/index.tsx
```diff
@@ -11,13 +16,33 @@ const AnsiiRow = forwardRef<any, BaseRowProps>((rowProps, ref) => {
...
 });
 
+interface ProcessedAnsiiRowProps {
+  lineContent: string;
+}
+const ProcessedAnsiiRow: React.FC<ProcessedAnsiiRowProps> = ({
+  lineContent,
+}) => {
+  const memoizedLogLine = useMemo(() => {
+    if (!lineContent) return null;
```
minnakt, line 35: methinks you can delete this since it seems like it should be guaranteed to not be undefined

```suggestion

```

minnakt (approved):

PR #37, opened 2022-09-12
EVG-17535: Create CollapsedRow component

minnakt:
EVG-17535

### Description 
This PR creates the `CollapsedRow` component. You can view the component either in the stories or by uncommenting the line in `LogWindow/index.tsx` (for regular logs, not for uploaded logs). Also this PR is not actually that big, the snapshots just added a lot of lines.

### Note
I changed the design of this component slightly from the MVP because I felt like it didn't contain all the information a user would want compared to legacy Lobster.

### Screenshots
Screen Shot 2022-09-12 at 10 09 55 AM

### Testing 
- Added story to `AnsiiRow` and `ResmokeRow`
- Added `CollapsedRow.test.tsx`


src/components/LogRow/CollapsedRow/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const RowContainer = styled.div`
+  display: flex;
+  align-items: center;
+  overflow-y: hidden;
+
+  height: ${size.m} !important;
+  padding-left: ${size.s};
+  background-color: ${gray.light2};
+`;
```
khelif96, line 45: It looks like if there are more than a few rows this starts to break rendering. 
image

khelif96: I replicated this by implementing a rudimentary search functionality. 
Inside of LogWindow.tsx. i replaced `processedLogLines` with 
```
  const [search] = useQueryParam(QueryParams.Search, "");

  const processedLogLines = useMemo(() => {
    if (!search) {
      return [...Array(lineCount).keys()];
    }
    const regex = new RegExp(search, "i");
    const matchingSearches = [];
    for (let i = 0; i < lineCount; i++) {
      const line = getLine(i) || "";
      if (regex.test(line)) {
        matchingSearches.push(i);
      } else if (Array.isArray(matchingSearches[matchingSearches.length - 1])) {
        // @ts-expect-error
        matchingSearches[matchingSearches.length - 1].push(i);
      } else {
        matchingSearches.push([i]);
      }
    }
    return matchingSearches;
  }, [search, lineCount, getLine]);
  ```

**Review by khelif96 — CHANGES_REQUESTED** [review]

File: src/components/LogRow/AnsiiRow/index.tsx [hunk 2]
```diff
@@ -13,13 +14,19 @@ const AnsiiRow = forwardRef<any, BaseRowProps>((rowProps, ref) => {
...
+  }
+
+  const lineContent = getLine(line);
+  if (!lineContent) {
+    return null;
+  }
   return (
-    <BaseRow wrap={wrap} {...listRowProps} ref={ref}>
+    <BaseRow wrap={wrap} {...listRowProps} ref={ref} lineNumber={line}>
       <ProcessedAnsiiRow lineContent={lineContent} />
```
minnakt, line 30: Words cannot express how much this upsets me

src/components/LogRow/ResmokeRow/index.tsx
```diff
@@ -1,19 +1,20 @@
...
-  const lineIndex = Array.isArray(line) ? line[0] : line;
-  return (
-    <BaseRow ref={ref} wrap={wrap} {...listRowProps}>
-      <span data-cy="resmoke-row">{getLine(lineIndex)}</span>
+  return Array.isArray(line) ? (
+    <CollapsedRow ref={ref} {...listRowProps} numCollapsed={line.length} />
+  ) : (
+    <BaseRow ref={ref} wrap={wrap} {...listRowProps} lineNumber={line}>
+      <span data-cy="resmoke-row">{getLine(line)}</span>
     </BaseRow>
```
khelif96, line 18: image
I still see the weird styling for resmoke rows.

khelif96: I think it makes sense to align the lines so the buttons line up

src/components/LogRow/AnsiiRow/index.tsx
```diff
@@ -13,13 +14,19 @@ const AnsiiRow = forwardRef<any, BaseRowProps>((rowProps, ref) => {
...
   const line = processedLines[index];
-  // TODO: EVG-17535
-  // This should be replaced with a collapsible component
-  const lineIndex = Array.isArray(line) ? line[0] : line;
-  const lineContent = getLine(lineIndex);
-  if (!lineContent) return null;
+
+  if (Array.isArray(line)) {
+    return (
+      <CollapsedRow ref={ref} {...listRowProps} numCollapsed={line.length} />
```
khelif96, line 20: Same goes here
image

khelif96 (changes requested):

src/components/LogRow/AnsiiRow/index.tsx
khelif96: Its not an issue when `wrap=true` so that might have something to do with it. 
image

khelif96 (approved):

src/components/LogPane/index.tsx
```diff
@@ -23,7 +25,7 @@ const LogPane: React.FC<LogPaneProps> = ({
     cache.clearAll();
     listRef.current?.recomputeRowHeights();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [wrap]);
+  }, [wrap, search]);
```
khelif96, line 28: Although this should technically be filters not search. cc @minnakt

khelif96: Can you ensure we update this in your follow up pr.

PR #39, opened 2022-09-13
EVG-17539 Resmoke line processing

khelif96:
EVG-17539

### Description 
Introduces Resmoke Line processing of log lines as they are ingested. 
Opting to use regex to parse the log lines and the subsequent json instead of JSON.parse(). From what i can tell it's slightly more performant. Also avoids some data corruption caused by transforming the log lines into native js primitives since some mongodb bson types are incompatible with javascript. See https://jira.mongodb.org/browse/EVG-14065

Would love a spot check by any reviewers comparing to legacy lobster.
### Screenshots

https://user-images.githubusercontent.com/4605522/190646006-0dc48a01-e9d1-401b-9699-fc97f86387c3.mov

### Testing 
Wrote unit tests that compare my log file output with those of legacy lobster


src/context/LogContext/LogContext.test.tsx
```diff
@@ -16,11 +17,30 @@ describe("useLogContext", () => {
...
+        [
+          `[j0:s0:n1] {"t":{"$date":"2022-09-13T16:57:46.852+00:00"},"s":"D2", "c":"REPL_HB",  "id":4615670, "ctx":"ReplCoord-1","msg":"Sending heartbeat","attr":{"requestId":3705,"target":"localhost:20003","heartbeatObj":{"replSetHeartbeat":"shard-rs0","configVersion":5,"configTerm":3,"hbv":1,"from":"localhost:20004","fromId":1,"term":3,"primaryId":1}}}`,
+          `[j0:s0] {"t":{"$date":"2022-09-13T16:57:46.855+00:00"},"s":"I",  "c":"-",        "id":20883,   "ctx":"conn188","msg":"Interrupted operation as its client disconnected","attr":{"opId":6047}}`,
+        ],
+        LogTypes.RESMOKE_LOGS
+      );
+    });
+    expect(result.current.logLines).toStrictEqual([
+      `[j0:s0:n1] | 2022-09-13T16:57:46.852+00:00 D2 REPL_HB  4615670 [ReplCoord-1] "Sending heartbeat","attr":{"requestId":3705,"target":"localhost:20003","heartbeatObj":{"replSetHeartbeat":"shard-rs0","configVersion":5,"configTerm":3,"hbv":1,"from":"localhost:20004","fromId":1,"term":3,"primaryId":1}}`,
+      `[j0:s0] | 2022-09-13T16:57:46.855+00:00 I  -        20883   [conn188] "Interrupted operation as its client disconnected","attr":{"opId":6047}`,
```
minnakt, line 41: I wonder if there is a way we can test this without the exact string whitespaces... Maybe we can trim and compare? it makes me kind of sad that we'd need to update many places if we ever decide to change the `padEnd` values 😭

src/utils/resmoke/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  const attr = getAttributes(json);
+
+  const output = `${resmokeFunction} ${port ?? ""}| ${timeStamp ?? ""} ${
+    shellPrefix ?? ""
+  } ${configSrv ?? ""} ${id ?? ""} [${ctx ?? ""}] "${msg ?? ""}"${
+    attr ? `,${attr}` : ""
+  }`;
+  return output;
+};
+
```
minnakt, line 43: Do we need to extract whatever these `truncated` and `size` values are, or do they seem like deprecated stuff?

https://github.com/evergreen-ci/lobster/blob/9039b4a104689afc5c86d46fd869397f8deea239/src/selectors/lines/transforms.js#L26-L29

minnakt (changes requested):

src/utils/resmoke/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  const json = getJSONString(line);
+  const port = getPort(line);
+  if (!json) {
+    return line;
+  }
+
+  const timeStamp = getTimeStamp(json);
+  const shellPrefix = getShellPrefix(json)?.padEnd(2, " ");
+  const configSrv = getConfigServer(json)?.padEnd(8, " ");
+  const id = getId(json)?.padEnd(7, " ");
```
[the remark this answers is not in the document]
khelif96: 🤷 now it is

src/context/LogContext/LogContext.test.tsx
khelif96: I think the whitespaces are part of the log. I know server has some regexp that they use to find errors so i fear changing the tests would break things.

src/utils/resmoke/index.ts
khelif96: I couldn't actually find any logs with these fields so i'm assuming its possibly something legacy? I think it might be ok to ignore them for now and if we realize we need them it should be easy to add them back. It's difficult to craft a regex for them if i don't have a reference.

minnakt (approved):

src/utils/resmoke/helpers.ts
```diff
@@ -0,0 +1,165 @@
...
+ * @example "c":"REPL"
+ */
+const configSrvRegex = /"c":"([a-zA-Z-_]+)"/;
+/** `getConfigServer` returns the config server associated with a resmoke function this is found in the resmoke json  */
+const getConfigServer = (line: string) => {
+  const configSrv = line.match(configSrvRegex)?.[1];
+  return configSrv;
+};
+
+// regex that matches  "c":"NETWORK" or "c":"STORAGE" or "c":"COMMAND" or "c":"REPL" and nothing else
```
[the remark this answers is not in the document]
khelif96: oops

PR #40, opened 2022-09-16
EVG-17914 Add code splitting to produce smaller bundles

khelif96:
EVG-17914

### Description 
Our main bundle has gotten kind of big due to some dependencies that most users may not ever interact with. (LogUploader , Logs with AnsiiRow) So i thought it would make sense to split out those deps into their own bundles.

### Screenshots
bundle before:
image

bundle after:
image

### DIFF
--- a/src/components/LogRow/AnsiiRow/index.tsx
+++ b/src/components/LogRow/AnsiiRow/index.tsx
@@ -47,4 +47,4 @@ const ProcessedAnsiiRow: React.FC<ProcessedAnsiiRowProps> = ({
 
 AnsiiRow.displayName = "AnsiiRow";
 
-export { AnsiiRow };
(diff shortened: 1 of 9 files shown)


src/components/LogRow/AnsiiRow/index.tsx
```diff
@@ -47,4 +47,4 @@ const ProcessedAnsiiRow: React.FC<ProcessedAnsiiRowProps> = ({
 
 AnsiiRow.displayName = "AnsiiRow";
 
-export { AnsiiRow };
+export default AnsiiRow;
```
minnakt, line 50: Should `ResmokeRow` have the same treatment? (`export default`)

minnakt (approved):

PR #42, opened 2022-09-19
EVG-17525: Implement basic filtering

minnakt:
EVG-17525

### Description 
This PR implements basic filtering (assumes no case sensitivity, no visibility off, and exact match as opposed to inverse match).

### Screenshots
Screen Shot 2022-09-19 at 12 43 30 PM

### Testing 
- Added unit test `filter.test.ts`
- Added Cypress tests for each of the log views


src/components/LogPane/index.tsx
```diff
@@ -25,7 +25,7 @@ const LogPane: React.FC<LogPaneProps> = ({
     cache.clearAll();
     listRef.current?.recomputeRowHeights();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [wrap, filters]);
+  }, [wrap, `${filters}`]);
```
minnakt, line 28: This hook was being called too much because every `filters` array was being interpreted as a new reference :\

A way to get around this is to stringify the array so that we'll run the hook only when the string changes

src/utils/filter/index.ts
```diff
@@ -0,0 +1,70 @@
...
+export const filterLogs = (
+  logLines: string[],
+  filters: string[],
+  bookmarks: number[],
+  selectedLine: number | undefined,
+  filterLogic: FilterLogic
+): (number | number[])[] => {
+  if (filters.length === 0) {
+    return [...Array(logLines.length).keys()];
+  }
```
khelif96, line 42: I wonder if theres a way to do this without creating a new array save some memory...?

khelif96 (approved):
Great job!

PR #43, opened 2022-09-20
EVG-17619 Setup logkeeper instance

khelif96:
EVG-17619

### Description 

Tries to start logkeeper on CI

There is a logkeeper  file in the task artifacts showing logkeepers log output.


minnakt (approved):
yaml looks good to me though I don't know much about logkeeper 🤷‍♂️

PR #44, opened 2022-09-22
EVG-17539 Adjust resmoke function processing

khelif96:
EVG-17539

### Description 
Upon further investigation of resmoke lines i discovered mongod does not always separate the resmoke function with a pipe. So we need to add some further processing to separate the function.
### Screenshots
example bad lines
image

### DIFF
--- a/src/utils/resmoke/helpers.test.ts
+++ b/src/utils/resmoke/helpers.test.ts
@@ -130,21 +130,17 @@ describe("resmoke/helpers", () => {
       expect(getResmokeFunction("hello")).toBeUndefined();
     });
     it("returns undefined if a line has a resmoke function but no json", () => {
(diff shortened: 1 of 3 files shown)


minnakt (changes requested):

src/utils/resmoke/helpers.test.ts
```diff
@@ -129,11 +129,19 @@ describe("resmoke/helpers", () => {
     it("handles strings without resmoke", () => {
       expect(getResmokeFunction("hello")).toBeUndefined();
     });
-    it("handles strings with resmoke", () => {
+    it("returns undefined if a line has a resmoke function but no json", () => {
```
minnakt, line 132: I think you might need to change this test descriptor text now

minnakt (approved):

PR #46, opened 2022-09-23
EVG-17663: Create Toast context

minnakt:
[EVG-17663](https://jira.mongodb.org/browse/EVG-17663)
[EVG-17664](https://jira.mongodb.org/browse/EVG-17664) (I think this ticket may also be covered by this PR)

### Description 
This PR adds the `ToastProvider` to Parsley. It is pretty much identical to the one in Spruce, though I changed some variable names to make it easier to understand.

The one big difference is in `toast.test.tsx`. I took a different approach to test the hook, let me know what you think 🤔

### Note
Whenever possible, we should be using `userEvent` rather than `fireEvent`. Only use `fireEvent` when it's absolutely necessary to test something.

Apparently we haven't been using the recommended syntax for `userEvent`, which is documented [here](https://testing-library.com/docs/user-event/setup/). This means that going forward, we should be doing something like:

```
const user = userEvent.setup()
await user.click(myElement) // always await a user event!
```

### Screenshots
Screen Shot 2022-09-23 at 4 06 28 PM

### Testing 
- Added `toast.test.tsx`
- Modified some tests to use `userEvent` rather than `fireEvent`


src/context/toast/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+        }),
+      info: (message = "", closable = true, options = {}) =>
+        dispatchToast({
+          variant: mapToastToLeafyGreenVariant.info,
+          message,
+          closable,
+          ...defaultOptions,
+          ...options,
+        }),
+      progress: (message = "", closable = true, options = {}) =>
```
khelif96, line 114: I think we should type the optional progress param

src/constants/toast.ts
```diff
@@ -0,0 +1,23 @@
+import { Variant } from "@leafygreen-ui/toast";
```
khelif96, line 1: I wonder if it makes sense to put all of the toast code in the toasts folder. It seems pretty fragmented being in 3 locations especially if only toast is going to consume it.

khelif96:
Good catch on the userEvent stuff!

src/context/toast/index.tsx
minnakt: i didnt even see that, that's a cool prop 🤯

khelif96 (approved):

PR #48, opened 2022-09-26
EVG-17908: Implement filters drawer

minnakt:
EVG-17908

### Description 
This PR removes the Filters Overlay and replaces it with the Filters Drawer.

Other things:
* I added the MongoDB file icon to the `Subheader`.
* I removed the `calc` CSS we were using on various components' widths and heights and replaced it with `flex-grow` so that we don't have to manually calculate things anymore. I modified CSS across a few different files to accommodate for this.

### Screenshots
Screen Shot 2022-09-26 at 5 08 28 PM

### Testing 
* Added tests for `Filter.test.tsx`
* Added tests for `FiltersDrawer.test.tsx`
* Added e2e tests though they will need to be modified following [EVG-17515](https://jira.mongodb.org/browse/EVG-17515)


src/components/FiltersDrawer/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { size, zIndex } from "constants/tokens";
+import { useQueryParam } from "hooks/useQueryParam";
+import Filter from "./Filter";
+
+interface FiltersDrawerProps {
+  ["data-cy"]?: string;
+}
+
+const FiltersDrawer: React.FC<FiltersDrawerProps> = ({ "data-cy": dataCy }) => {
+  const [collapsed, setCollapsed] = useState(true);
```
sophstad, line 17: I could see some benefit to defaulting collapsed to `false`... may help prevent a few "how do I apply filters" questions from users 😆

src/components/FiltersDrawer/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <StyledSideNav
+      aria-label="Filters Side Nav"
+      collapsed={collapsed}
+      data-cy={dataCy}
+      setCollapsed={setCollapsed}
+      widthOverride={250}
+    >
+      <PaddedContainer>
+        <StyledSideNavGroup
+          glyph={<Icon glyph="Filter" />}
```
sophstad, line 50: It'd be sweet if this icon appeared inactive (i.e. grey) when 0 filters are applied... but I'm looking at the LG component right now and it doesn't appear so easy 😿

src/components/FiltersDrawer/Filter/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+            <Button onClick={() => resetEditState()} size="xsmall">
+              Cancel
+            </Button>
+          </ButtonWrapper>
+        </>
+      ) : (
+        <FilterName>{filterName}</FilterName>
+      )}
+
+      <StyledSegmentedControl
```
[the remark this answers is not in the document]
minnakt: I added the labels, though I had to use some not-great CSS and I have a feeling the filters are starting to look crowded. Maybe we can ask more about it at design sync!

src/components/FiltersDrawer/index.tsx
minnakt: I changed it to use a cookie!

src/components/FiltersDrawer/index.tsx
minnakt: when there's a will, there's a way 😁 if you would like me to tweak the colors, let me know!

src/components/FiltersDrawer/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+        setCollapsed(collapse);
+        Cookie.set(HAS_OPENED_DRAWER, "true", { expires: 365 });
+      }}
+      widthOverride={270}
+    >
+      <PaddedContainer>
+        <StyledSideNavGroup
+          glyph={
+            <Icon
+              fill={filters.length ? green.dark2 : gray.base}
```
sophstad, line 62: Damn—that looks perfect!! Thanks 🙏

src/components/FiltersDrawer/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+  display: flex;
+  flex-direction: column;
+  padding: 0 ${size.xs};
+`;
+
+// @ts-expect-error
+const StyledSideNavGroup = styled(SideNavGroup)`
+  > div {
+    padding: 0;
+  }
```
sophstad, line 109: A little extra sauce 😩
```suggestion
  }
  & svg {
    transition: color 0.3s ease;
  }
```

src/components/FiltersDrawer/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  return (
+    <StyledSideNav
+      aria-label="Filters Side Nav"
+      collapsed={collapsed}
+      data-cy={dataCy}
+      setCollapsed={(collapse: React.SetStateAction<boolean>) => {
+        setCollapsed(collapse);
+        Cookie.set(HAS_OPENED_DRAWER, "true", { expires: 365 });
```
sophstad, line 54: Instead of using this as a flag, why not open the drawer initially and then use this to preserve whatever state the user leaves the drawer in? Seems like that may lead to the most consistent experience.

src/components/FiltersDrawer/Filter/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <>
+          <StyledTextInput
+            aria-label="Edit Filter Name"
+            data-cy="edit-filter-name"
+            onChange={(e) => setNewFilter(e.target.value)}
+            placeholder="New filter definition"
+            sizeVariant="small"
+            spellCheck={false}
+            type="search"
+            value={newFilter}
```
sophstad, line 93: Just flagging based on design review today we should show the existing filter value here for users to edit!

src/components/FiltersDrawer/Filter/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+          <ButtonWrapper>
+            <Button
+              onClick={() => {
+                editFilter(filterName, newFilter);
+                resetEditState();
+              }}
+              size="xsmall"
+              variant={Variant.PrimaryOutline}
+            >
+              OK
```
sophstad, line 104: Based on LG patterns in the confirmation modal, I think having OK to the right here may be more intuitive

src/components/FiltersDrawer/Filter/index.tsx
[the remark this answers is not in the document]
sophstad: Nice, I think you did the best we can do with the layout here 😄

src/components/FiltersDrawer/Filter/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+    setNewFilter("");
+  };
+
+  return (
+    <FilterRow data-cy={dataCy}>
+      <FilterLabel>
+        <Badge>FILTER</Badge>
+        <FilterIcons>
+          <IconButton
+            aria-label="Edit filter button"
```
sophstad, line 60: [nit] Aria label text doesn't need to announce the role of the element, it can just describe its function briefly. There are some good guidelines [here](https://www.aditus.io/aria/aria-label/#when-to-use)!
```suggestion
            aria-label="Edit filter"
```

src/components/FiltersDrawer/Filter/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+            onClick={() => setIsEditing(true)}
+          >
+            <Icon fill={gray.base} glyph="Edit" />
+          </IconButton>
+          <IconButton
+            aria-label="Visibility filter button"
```
sophstad, line 66: [nit] Here we could toggle the text
```suggestion
            aria-label={isVisible ? "Hide filter" : "Show filter"}
```

src/components/FiltersDrawer/Filter/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+            onClick={() => setIsVisible(!isVisible)}
+          >
+            <Icon
+              fill={gray.base}
+              glyph={isVisible ? "Visibility" : "ClosedEye"}
+            />
+          </IconButton>
+          <IconButton
+            aria-label="Delete filter button"
```
sophstad, line 75:
```suggestion
            aria-label="Delete filter"
```

src/components/FiltersDrawer/index.tsx
minnakt: I decided to leave this out, only because it's a small thing that could go unnoticed and I want to minimize editing the internal CSS of LeafyGreen components as much as possible 🙏

src/components/FiltersDrawer/index.tsx
minnakt: if it's okay I would like to leave it like this for now 🥺 It should be easy to switch later!

sophstad (approved):

PR #49, opened 2022-09-27
EVG-17528 Search functionality  

khelif96:
EVG-17528

### Description 
Sorry about the longer PR but it was necessary.

I did a little bit of refactoring in an effort to make the state management a little more manageable and to avoid passing things up and down the react tree. 
I moved `processedLogLines` to the app context so that it can be accessed by the search functionality of the app. (This is accessed in the navbar so it didn't make sense to have it live in `LogWindow` 
To support that i moved the router hooks to the context which meant moving the order of the app contexts around.
This ended up causing me to require a router context for many stories that ingest a log context hopefully that explains some of the seemingly random updates.

### Screenshots
This video cheats a little because i'm using keyboard shortcuts for interactions but didn't include them in this pr. But the general idea of the feature should make sense. 

https://user-images.githubusercontent.com/4605522/193669769-4a708d80-f28c-44c9-a62d-767ac4d0ae3b.mov

### Testing 
Wrote unit and e2e tests.
Resmoke row tests incoming but they resemble ansii.


src/utils/collapsedRow/index.ts
```diff
@@ -0,0 +1,12 @@
+import { ProcessedLogLine } from "types/logs";
+
+/**
+ * isCollapsedRow determines if a row is a collapsed row
+ * although it is a simple function its purpose is to make the code more readable
+ * @param logLine ProcessedLogLine
+ * @returns boolean true if the row is a collapsed row
+ */
+const isCollapsedRow = (logLine: ProcessedLogLine): logLine is number[] =>
```
khelif96, line 9: This fancy `logline is number[]` syntax supports typescripts control flow analysis

src/context/LogContext/index.tsx
```diff
@@ -31,42 +52,135 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
+      lowerRange,
+      processedLogLines,
+      getLine,
+    ]
   );
+
+  // Handle case sensitivity changes
+  useEffect(() => {
+    dispatch({ type: "TOGGLE_CASE_SENSITIVE", caseSensitive });
+  }, [caseSensitive, dispatch]);
```
minnakt, line 114: I wonder if it would be better to have this function accessible from `LogContext()` and called directly in `CaseSensitiveToggle` on change

src/utils/renderHtml/index.tsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+import parse, {
+  Element,
+  HTMLReactParserOptions,
+  domToReact,
+} from "html-react-parser";
+
+/**
+ * renderHtml takes a string and converts it into an array of domnodes and
```
minnakt, line 9: I'm curious if you looked into [`react-highlight-words`](https://github.com/bvaughn/react-highlight-words)? It's also written by bvaughn (he's everywhere... 🤔). It could be a simpler approach since we're only using this for highlighting words

minnakt (changes requested):
Just some comments to start, since this a big PR 😅

src/utils/renderHtml/index.tsx
khelif96: I have! But this actually does more than just highlight the words it also handles rendering the ansii transformed html string into dom elements that are generated by ansii up. I figured might as well just extend the same library we are using elsewhere to also do the syntax highlighting instead of pulling in yet another library.

src/components/NavBar/SearchResults/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+        </Button>
+      </>
+    )}
+  </SearchContainer>
+);
+
+const SearchContainer = styled.div`
+  align-items: center;
+  display: flex;
+`;
```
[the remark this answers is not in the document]
khelif96: Wow good one i didn't know about gap!

src/components/SearchBar/index.tsx
```diff
@@ -15,27 +16,45 @@ interface SearchBarProps {
   validatorMessage?: string;
   className?: string;
   onSubmit?: (selected: string, value: string) => void;
+  onChange?: (selected: string, value: string) => void;
+  shouldClearOnSubmit?: (selected: string) => boolean;
 }
 
 const SearchBar: React.FC<SearchBarProps> = ({
```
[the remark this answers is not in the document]
khelif96: Good catch!

src/context/LogContext/index.tsx
khelif96: Great idea!

src/context/LogContext/index.tsx
```diff
@@ -31,42 +52,135 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [state.logs.length > 0, selectedLine]);
+
+  // If the search term changes, scroll to the first match
+  useEffect(() => {
+    if (state.searchState.searchIndex !== undefined) {
+      scrollToLine(searchResults[state.searchState.searchIndex]);
+    }
+  }, [scrollToLine, searchResults, state.searchState.searchIndex]);
```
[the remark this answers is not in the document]
khelif96: I couldn't replicate this flicker but great idea! It would reduce a render either way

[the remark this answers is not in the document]
minnakt: its my Macbook pro debuff 🙁

src/context/LogContext/index.tsx
```diff
@@ -31,42 +53,128 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [dispatch, `${searchResults}`, state.searchState.searchTerm]);
+
+  // If the selected line changes, scroll to it
+  useEffect(() => {
+    if (selectedLine) {
+      scrollToLine(selectedLine);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [!!state.logs.length, selectedLine]);
```
minnakt, line 126: Rather than use an effect, I think I would prefer to pass `scrollToLine` through the `RowRenderer` and call it directly in `BaseRow`

src/context/LogContext/index.tsx
```diff
@@ -31,42 +53,128 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
   );
+
+  // Handle search result changes
+  useEffect(() => {
+    dispatch({
+      type: "SET_MATCH_COUNT",
+      matchCount: searchResults.length,
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [dispatch, `${searchResults}`, state.searchState.searchTerm]);
```
minnakt, line 118: Sorry for being a `useEffect` hater... but I want to believe that we can somehow refactor this to not have to use an effect. Maybe not the best option, but it could definitely be moved to the `useMemo` for starters

src/components/DetailsOverlay/Toggles/CaseSensitiveToggle/index.tsx
```diff
@@ -1,18 +1,14 @@
...
-import { useQueryParam } from "hooks/useQueryParam";
+import { useLogContext } from "context/LogContext";
 import BaseToggle from "../BaseToggle";
 
 const CaseSensitiveToggle: React.FC = () => {
-  const [caseSensitive, setCaseSensitive] = useQueryParam(
-    QueryParams.CaseSensitive,
-    false
-  );
+  const { searchState, setCaseSensitive } = useLogContext();
```
minnakt, line 5: Is it intentional that this toggle no longer updates the URL params (i.e. do we want case sensitivity to be shareable for search)?

src/components/NavBar/index.tsx
```diff
@@ -5,31 +5,46 @@ import Icon from "components/Icon";
...
   };
 
+  const handleOnChange = (selected: string, value: string) => {
+    if (selected === SearchBarActions.Search) {
+      setSearch(value);
+    }
+  };
+
+  const shouldClearOnSubmit = (selected: string) =>
+    selected === SearchBarActions.Filter;
```
minnakt, line 46: I feel like we don't really need to have this as a function / props. It might be better and easier to read if we just define it directly in `Searchbar` as 

```
const shouldClearOnSubmit = selected === SearchBarActions.Filter
```

I don't think it should affect the additional Jest tests you wrote for `SearchBar` very much at all

src/utils/renderHtml/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+      if (domNode instanceof Element) {
+        if (options.transform && options.transform[domNode.name]) {
+          const SwapComponent = options.transform[domNode.name];
+          // SwapComponent is just what ever component we want to return from the transform object
+          const extraProps =
+            options.preserveAttributes &&
+            options.preserveAttributes.includes(domNode.name)
+              ? domNode.attribs
+              : {};
+          if (SwapComponent) {
```
minnakt, line 42: this might be a silly question, but why do we need to check for `if (SwapComponent)` when it's defined as `options.transform[domNode.name]` which is checked on `L34`?

src/components/SearchBar/index.tsx
```diff
@@ -77,13 +105,13 @@ const SearchBar: React.FC<SearchBarProps> = ({
...
         }
-        onChange={(e) => setInput(e.target.value)}
+        onChange={(e) => handleOnChange(e.target.value)}
         onKeyPress={(e: KeyboardEvent<HTMLInputElement>) =>
           e.key === "Enter" && handleOnSubmit()
         }
         placeholder="optional, regexp to search"
         spellCheck={false}
-        type="search"
+        type={SearchBarActions.Search}
```
minnakt, line 114: I don't think we should type this with our enum since it's a native property of `input`.

If you really want you could probably type it using `import { TextInputType } from "@leafygreen-ui/text-input"` but that might be a little overkill

src/components/NavBar/SearchResults/SearchCount.tsx
```diff
@@ -0,0 +1,27 @@
...
+  currentSearchIndex?: number;
+}
+const SearchCount: React.FC<SearchCountProps> = ({
+  currentSearchIndex,
+  matchingSearchCount,
+}) => (
+  <StyledBody data-cy="search-count" hasMatches={matchingSearchCount !== 0}>
+    {matchingSearchCount !== 0
+      ? `${currentSearchIndex}/${matchingSearchCount}`
+      : "Not Found"}
```
minnakt, line 18:
```suggestion
      : "No Matches"}
```

src/components/Highlight/index.tsx
```diff
@@ -0,0 +1,14 @@
+import styled from "@emotion/styled";
+import { palette } from "@leafygreen-ui/palette";
+
+const { red } = palette;
+
+const Highlight = styled.mark<{ "data-cy"?: string }>`
+  background-color: ${red.light2};
```
minnakt, line 7: nit. this could use some opacity. this is kind of sad but I think we can't use the color tokens with `rgba` so we would need to use `#rrggbbaa` format

here's a list of transparency codes: [link](https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4)

src/utils/renderHtml/index.tsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
```
minnakt, line 1:
```suggestion

```

src/utils/renderHtml/renderHtml.test.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
```
minnakt, line 1:
```suggestion

```

src/components/SearchBar/index.tsx
```diff
@@ -15,42 +17,68 @@ interface SearchBarProps {
...
   disabled = false,
+  onChange,
+  onSubmit,
+  shouldClearOnSubmit = () => true,
   validator = () => true,
   validatorMessage = "Invalid Input",
-  className,
-  onSubmit,
 }) => {
   const [input, setInput] = useState("");
```
minnakt, line 33: I wonder if we should maintain two different inputs for Filter and Search... 🤔

for example, if you do the following:
1. apply a search
2. switch to filter and apply a filter
3. go back to search

the search term is gone from the input but the search is still being applied

src/context/LogContext/index.tsx
```diff
@@ -31,42 +53,128 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
+    () =>
+      // search through processedLoglines
+      // return the line number of the first match
+      // if no match, return undefined
+      state.searchState.searchTerm
+        ? searchLogs({
+            searchRegex: state.searchState.searchTerm,
+            processedLogLines,
+            upperBound: upperRange,
+            lowerBound: lowerRange,
```
minnakt, line 98: When it comes to the bounds, legacy Lobster has a slightly different behavior.

this PR (even if term is out of bounds, highlights it):
Screen Shot 2022-10-06 at 9 43 49 AM

legacy Lobster (if term is out of bounds, there is no highlight):
Screen Shot 2022-10-06 at 9 44 09 AM

minnakt (changes requested):
this is looking really good!

src/components/DetailsOverlay/Toggles/CaseSensitiveToggle/index.tsx
khelif96: Yes it is i think it shouldn't be we don't share search results and since case sensitivity only affects search it doesn't make sense to share it.

src/context/LogContext/index.tsx
khelif96: Great catch!

src/components/SearchBar/index.tsx
khelif96: Ooops that def wasn't intentional

src/components/SearchBar/index.tsx
khelif96: I replicated legacy behavior we now clear the search after you apply a filter.

src/components/Highlight/index.tsx
khelif96: This will be handled in 
https://jira.mongodb.org/browse/EVG-18030

minnakt (changes requested):
me watching the downfall of useEffect

evil-burns

src/components/LogRow/AnsiiRow/index.tsx
```diff
@@ -1,47 +1,80 @@
...
+    if (searchTerm) {
+      render = render.replace(searchTerm, `<mark>$&</mark>`);
+    }
+    return renderHtml(render, {
+      transform: {
+        // @ts-expect-error - This is expecting a react component but its an Emotion component which are virtually the same thing
+        mark: Highlight,
+      },
+    });
+  }, [lineContent, searchTerm]);
```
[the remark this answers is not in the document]
khelif96: done lmk if this works for you?

src/utils/collapsedRow/index.ts
```diff
@@ -0,0 +1,12 @@
+import { ProcessedLogLine } from "types/logs";
+
+/**
```
minnakt, line 3: this function might belong better in `LogRow/utils.ts`!

minnakt (approved):
thanks for all your work on this momentous task! 🧙‍♂️

src/utils/collapsedRow/index.ts
khelif96: I think given the fact that it is also used by the filter util method it should remain in the utils folder. Its strange to import a function from a component folder for use in a generic util.

PR #50, opened 2022-09-27
EVG-17931 Remove lazy loading for AnsiiRow

khelif96:
EVG-17931

### Description 
Cell Renderer seems to have some issues rendering lazy loaded components so for now i'm going to disable it.


minnakt (approved):
The PR title is a little long and confusing.

Let's simplify it to something like `Remove lazy loading for AnsiiRow`

PR #52, opened 2022-09-30
EVG_18007 test_utils now wraps element wrappers in router

khelif96:
EVG-18007

### Description 
Ran into this while writing tests for EVG-17528. Any elements that are passed into the wrapper prop aren't actually wrapped with a react router instance. This prevents us from using react router within a context.


minnakt (approved):
LGTM, though PR title should be `EVG-18007` rather than `EVG_18007`. I would also suggest referencing `renderWithRouterMatch` specifically rather than `test_utils` as a whole

PR #64, opened 2022-10-03
EVG-17665: Add 404 page

minnakt:
EVG-17665

### Description 
This PR adds the 404 page.

### Screenshots
Screen Shot 2022-10-03 at 4 52 05 PM

### Testing 
- Edited Cypress test


src/pages/index.tsx
```diff
@@ -2,6 +2,7 @@ import { Suspense, lazy } from "react";
 import { Navigate, Route, Routes } from "react-router-dom";
 import { LogTypes } from "constants/enums";
 import routes from "constants/routes";
+import NotFound from "./404";
```
khelif96, line 5: We should lazy load this since the graphic is pretty big we don't want to download it unless we need to.

PR #66, opened 2022-10-04
EVG-17534 Linkify subheader component

khelif96:
EVG-17534

### Description 
Subheader task link now links to the corresponding task page.
Adds a useFetch hook to make fetch requests for json. This is used to fetch the `task_id` and `execution` for resmoke logs.

### Testing 
e2e and unit tests


src/components/SubHeader/index.tsx
```diff
@@ -3,18 +3,24 @@ import { palette } from "@leafygreen-ui/palette";
...
 }
 const SubHeader: React.FC<SubHeaderProps> = ({ isUploadedLog }) => {
   const { logMetadata } = useLogContext();
-  const { fileName } = logMetadata || {};
+  const { fileName, taskID, execution } = logMetadata || {};
+  let taskLink = "";
+
+  if (taskID !== undefined && execution !== undefined) {
+    taskLink = getEvergreenTaskURL(taskID, execution);
+  }
```
minnakt, line 22: just a suggestion, it might be a matter of preference
```suggestion
  const taskLink =
    taskID !== undefined && execution !== undefined
      ? getEvergreenTaskURL(taskID, execution)
      : "";
```

src/hooks/useFetch/index.ts
```diff
@@ -0,0 +1,60 @@
...
+  const [data, setData] = useState<T | null>(null);
+  const [error, setError] = useState<string | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    leaveBreadcrumb("useFetch", { url }, "request");
+    const req = new Request(url, { method: "GET" });
+    const abortController = new AbortController();
+
+    if (!options?.skip) {
```
minnakt, line 27: it might be a little nicer if we deconstruct `skip` above
```suggestion
  const [isLoading, setIsLoading] = useState(false);
  const { skip = false } = options || {};

  useEffect(() => {
    leaveBreadcrumb("useFetch", { url }, "request");
    const req = new Request(url, { method: "GET" });
    const abortController = new AbortController();

    if (!skip) {
```

src/hooks/useFetch/index.ts
```diff
@@ -0,0 +1,60 @@
...
+      setIsLoading(true);
```
minnakt, line 28: hmm instead of doing `setIsLoading` here, it might be possible to set it on initialization of state on `L20` based on the value of `skip`

src/hooks/useFetch/index.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    return () => {
+      // Cancel the request if the component unmounts
+      abortController.abort();
+    };
+  }, [url, options?.skip]);
+  return { data, error, isLoading };
+};
+
+export { useFetch };
```
minnakt, line 60: wondering if we should explore combining `useFetch` and `useLogDownloader` into one thing... I see that there are some differences (e.g. `response.json()` vs `response.text()`), but the bulk format is pretty similar. Or maybe we should do something like `useJsonFetch` and `useTextFetch`???

src/types/api.ts
```diff
@@ -0,0 +1,9 @@
+type LogKeeperMetadata = {
```
minnakt, line 1: I always thought the `k` was lowercase, but ignore me if it's wrong
```suggestion
type LogkeeperMetadata = {
```

src/pages/LogView/LoadingPage/index.tsx
```diff
@@ -37,6 +39,12 @@ const LoadingPage: React.FC<LoadingPageProps> = ({ logType }) => {
   let url = "";
   let htmlLogURL = "";
   let jobLogsURL = "";
+  const { data: logkeeperMetadata } = useFetch<LogKeeperMetadata>(
+    getResmokeLogURL(buildID || "", { testID, metadata: true }),
```
minnakt, line 43: do we need the `|| "" `?
```suggestion
    getResmokeLogURL(buildID, { testID, metadata: true }),
```

src/pages/LogView/LoadingPage/index.tsx
```diff
@@ -46,6 +54,12 @@ const LoadingPage: React.FC<LoadingPageProps> = ({ logType }) => {
         url = getResmokeLogURL(buildID, { raw: true });
         htmlLogURL = getResmokeLogURL(buildID, { html: true });
       }
+      if (logkeeperMetadata) {
+        jobLogsURL = getSpruceJobLogsURL(
+          logkeeperMetadata.task_id,
+          String(logkeeperMetadata.execution || 0)
+        );
```
minnakt, line 61: maybe we should allow `execution` in this function to be either a string or a number!

minnakt (changes requested):

src/components/SubHeader/index.tsx
khelif96: Your preference is better :)

src/hooks/useFetch/index.ts
khelif96: This actually isn't true because if we start by skipping the query and then setting `skip=false` isLoading won't be set back to true. I added a test to assert this.

src/pages/LogView/LoadingPage/index.tsx
khelif96: We do since `buildID` is not always defined. We use the skip to avoid making the query with the broken url but we still need to pass in an empty string to appease the type guards in `getResmokeLogURL`

minnakt (approved):
tyyy! exciting stuff

src/hooks/useFetch/useFetch.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    expect(result.current.isLoading).toBe(false);
+    expect(result.current.data).toBeNull();
+    skip = false;
+    rerender();
+    expect(result.current.isLoading).toBe(true);
+    await waitForNextUpdate();
+    expect(result.current.isLoading).toBe(false);
+    // eslint-disable-next-line jest/max-expects
+    expect(result.current.data).toBe(jsonMessage);
+  });
```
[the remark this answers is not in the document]
khelif96: Good point fixed!

PR #68, opened 2022-10-05
EVG-17542: Implement copy functionality for JIRA button

minnakt:
EVG-17542

### Description 
This PR adds the JIRA formatting functionality to the JIRA button in the Details Overlay panel.

### Screenshots

Screen Shot 2022-10-05 at 10 22 40 AM

### Testing 
- Added tests in `string.test.ts`
- Added tests in `ButtonRow.test.tsx`
- I copy pasted it into a JIRA ticket (you can try it on EVG-17542 if you would like)

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -10,6 +10,18 @@ Cypress.Commands.add("dataCy", (value: string) => {
   cy.get(`[data-cy=${value}]`);
 });
 
+Cypress.Commands.add("enableClipboard", () => {
+  cy.wrap(
+    Cypress.automation("remote:debugger:protocol", {
+      command: "Browser.grantPermissions",
+      params: {
+        permissions: ["clipboardReadWrite", "clipboardSanitizedWrite"],
+        origin: window.location.origin,
+      },
+    })
+  );
+});
+
 Cypress.Commands.add(
   "isContainedInViewport",
   { prevSubject: true },
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -10,6 +10,7 @@ Cypress.Commands.add("dataCy", (value: string) => {
   cy.get(`[data-cy=${value}]`);
 });
 
+// Source: https://stackoverflow.com/questions/60174546/how-grant-cypress-test-application-some-permissions
 Cypress.Commands.add("enableClipboard", () => {
   cy.wrap(
     Cypress.automation("remote:debugger:protocol", {
--- a/src/components/DetailsOverlay/ButtonRow/index.tsx
+++ b/src/components/DetailsOverlay/ButtonRow/index.tsx
@@ -1,13 +1,50 @@
+import { useState } from "react";
+import styled from "@emotion/styled";
 import Button from "@leafygreen-ui/button";
+import Tooltip from "@leafygreen-ui/tooltip";
 import Icon from "components/Icon";
+import { QueryParams } from "constants/queryParams";
+import { useLogContext } from "context/LogContext";
+import { useQueryParam } from "hooks/useQueryParam";
+import { constructJiraString, copyToClipboard } from "utils/string";
 import { DetailRow } from "../styles";
 
-const ButtonRow: React.FC = () => (
-  <DetailRow>
-    <Button leftGlyph={<Icon glyph="Copy" />}>JIRA</Button>
-    <Button leftGlyph={<Icon glyph="Export" />}>RAW</Button>
-    <Button leftGlyph={<Icon glyph="Export" />}>HTML</Button>
-  </DetailRow>
-);
+const ButtonRow: React.FC = () => {
+  const [hasCopied, setHasCopied] = useState(false);
+
+  const [bookmarks] = useQueryParam<number[]>(QueryParams.Bookmarks, []);
+  const { logLines } = useLogContext();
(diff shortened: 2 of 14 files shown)


src/components/DetailsOverlay/ButtonRow/index.tsx
```diff
@@ -1,13 +1,52 @@
...
-    <Button leftGlyph={<Icon glyph="Export" />}>RAW</Button>
-    <Button leftGlyph={<Icon glyph="Export" />}>HTML</Button>
-  </DetailRow>
-);
+const ButtonRow: React.FC = () => {
+  const [hasCopied, setHasCopied] = useState(false);
+
+  const [bookmarks] = useQueryParam<number[]>(QueryParams.Bookmarks, []);
+  const { logLines } = useLogContext();
+  const jiraString = constructJiraString(bookmarks, logLines);
```
khelif96, line 17: As a minor perf improvement Maybe we shouldn't generate the string at render time. I think it would make sense to generate it only when the user clicks on the button.

src/components/DetailsOverlay/ButtonRow/index.tsx
```diff
@@ -1,13 +1,52 @@
...
+  return (
+    <DetailRow>
+      <StyledTooltip
+        align="top"
+        justify="middle"
+        trigger={
+          <div>
+            <Button
+              leftGlyph={<Icon glyph="Copy" />}
+              onClick={() => {
```
khelif96, line 28: Lets disable the button when there are no bookmarks.

src/utils/string/string.test.ts
```diff
@@ -0,0 +1,25 @@
...
+      },
+    });
+    copyToClipboard("copy text");
+    expect(navigator.clipboard.writeText).toHaveBeenCalledTimes(1);
+    expect(navigator.clipboard.writeText).toHaveBeenCalledWith("copy text");
+  });
+});
+
+describe("constructJiraString", () => {
+  it("should properly format text", () => {
```
khelif96, line 17: nit. Separate this out into more test cases. So this can also serve as documentation for how the function is supposed to work.

As an example
"should add an ellipsis between lines if they aren't adjacent to each other"
"should print adjacent lines side by side"
"should properly format text"
"should print nothing when there are no bookmarks"
"should handle out of bounds bookmarks"

khelif96 (commented):
Great job on this can we include a cypress test for resmoke and this ansii as well please!

cypress/support/commands.ts
```diff
@@ -10,6 +10,18 @@ Cypress.Commands.add("dataCy", (value: string) => {
...
+  cy.wrap(
+    Cypress.automation("remote:debugger:protocol", {
+      command: "Browser.grantPermissions",
+      params: {
+        permissions: ["clipboardReadWrite", "clipboardSanitizedWrite"],
+        origin: window.location.origin,
+      },
+    })
+  );
+});
```
khelif96, line 23: :( Can you at least include a comment to the source of this.

src/components/DetailsOverlay/ButtonRow/index.tsx
```diff
@@ -1,13 +1,55 @@
...
+      </Tooltip>
+      <Button leftGlyph={<Icon glyph="Export" />}>RAW</Button>
+      <Button leftGlyph={<Icon glyph="Export" />}>HTML</Button>
+    </DetailRow>
+  );
+};
+
+// We need to wrap the button in a dummy div because mouse events are not triggered on
+// disabled elements.
+const ButtonWrapper = styled.div``;
```
khelif96, line 53: Why use a styled div a simple div would suffice above.

src/pages/LogView/LoadingPage/index.tsx
```diff
@@ -61,7 +61,7 @@ const LoadingPage: React.FC<LoadingPageProps> = ({ logType }) => {
   const { data, error, isLoading } = useAxiosGet(url);
   useEffect(() => {
     if (data) {
-      ingestLines(data.split("\n"), logType);
+      ingestLines(data.trim().split("\n"), logType);
```
khelif96, line 64: trim removes spaces from both ends of the string. I fear this may result in us removing intentional blank lines if there are some at the head or tail of a log file.

src/utils/string/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  let jiraString = "{noformat}\n";
+
+  for (let i = 0; i < bookmarks.length; i++) {
+    const bookmarkLine = bookmarks[i];
+    const logText = getLine(bookmarkLine);
+
+    // If bookmarks is out of bounds, stop processing.
+    if (logText === undefined) {
+      jiraString += "{noformat}";
+      return jiraString;
```
khelif96, line 32: We can use the same return statement and logic at the end of the for loop if we just break out of the loop instead of returning and doing the string append.
```suggestion
     break;
```

khelif96 (approved):
Just a few more comments but apart from that LGTM!

src/pages/LogView/LoadingPage/index.tsx
minnakt: I swapped it to `trimEnd` instead, which I think is good enough for now.

The new line at the end of any file doesn't render in the virtualized list for some reason, so it needs to be chopped off

PR #69, opened 2022-10-10
EVG-17622: Add keyboard shortcuts

minnakt:
EVG-17622

### Description 
This PR adds a hook called `useKeyboardShortcut` which can be used to register shortcuts with individual components. 

In the code you'll see that I conflated the `Control` key with the `Meta` key; that is because Windows users use the `Control` key to execute keyboard commands, while Mac users use the `Command/Meta` key to execute keyboard commands.

As an example I added the following keyboard shortcuts:
- `Ctrl + F`: Puts focus on the searchbar input.
- `Ctrl + D`: Opens and closes the filters drawer. Is not activated when there is focus on the searchbar input.
- `Ctrl + S`: Switches between the Search/Filter mode. Is activated when there is focus on the searchbar input.

### Screenshots
TODO

### Testing 
- Added tests in `useKeyboardShortcut.test.ts`
- Added a test in `SearchBar.test.tsx`

### DIFF
--- a/src/components/FiltersDrawer/index.tsx
+++ b/src/components/FiltersDrawer/index.tsx
@@ -9,6 +9,7 @@ import Icon from "components/Icon";
 import { HAS_OPENED_DRAWER } from "constants/cookies";
 import { QueryParams } from "constants/queryParams";
 import { size, zIndex } from "constants/tokens";
+import { useKeyboardShortcut } from "hooks";
 import { useQueryParam } from "hooks/useQueryParam";
 import Filter from "./Filter";
 
@@ -22,6 +23,8 @@ const FiltersDrawer: React.FC<FiltersDrawerProps> = ({ "data-cy": dataCy }) => {
   const [collapsed, setCollapsed] = useState(
     Cookie.get(HAS_OPENED_DRAWER) === "true"
   );
+
+  useKeyboardShortcut(["Control", "d"], () => setCollapsed(!collapsed));
 
   const [filters, setFilters] = useQueryParam<string[]>(
     QueryParams.Filters,
--- a/src/components/SearchBar/SearchBar.test.tsx
+++ b/src/components/SearchBar/SearchBar.test.tsx
@@ -50,4 +50,14 @@ describe("searchbar", () => {
     expect(input).toHaveValue("test");
     expect(onSubmit).not.toHaveBeenCalled();
   });
+  it("pressing Control+F puts focus on the input", async () => {
+    const user = userEvent.setup();
+    render(<SearchBar onSubmit={jest.fn()} />);
+
+    const input = screen.getByDataCy("searchbar-input");
+    expect(input).not.toHaveFocus();
+
+    await user.keyboard("{Control>}{f}");
+    expect(input).toHaveFocus();
+  });
 });
--- a/src/components/SearchBar/index.tsx
+++ b/src/components/SearchBar/index.tsx
@@ -1,4 +1,4 @@
-import { KeyboardEvent, useState } from "react";
+import { KeyboardEvent, useRef, useState } from "react";
 import styled from "@emotion/styled";
 import IconButton from "@leafygreen-ui/icon-button";
 import { palette } from "@leafygreen-ui/palette";
@@ -7,6 +7,7 @@ import Icon from "components/Icon";
 import IconWithTooltip from "components/IconWithTooltip";
 import TextInputWithGlyph from "components/TextInputWithGlyph";
 import { zIndex } from "constants/tokens";
+import { useKeyboardShortcut } from "hooks";
 
 const { yellow } = palette;
 interface SearchBarProps {
@@ -26,6 +27,23 @@ const SearchBar: React.FC<SearchBarProps> = ({
 }) => {
   const [input, setInput] = useState("");
   const [selected, setSelected] = useState("search");
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useKeyboardShortcut(
+    ["Control", "f"],
+    () => {
+      if (inputRef.current) inputRef.current.focus();
+    },
+    disabled
+  );
+
+  useKeyboardShortcut(
+    ["Control", "s"],
+    () =>
+      selected === "search" ? setSelected("filter") : setSelected("search"),
+    disabled,
+    { overrideIgnore: true }
+  );
 
   const isValid = validator(input);
 
@@ -63,6 +81,7 @@ const SearchBar: React.FC<SearchBarProps> = ({
             <IconButton
               aria-label="Select plus button"
               data-cy="searchbar-submit"
+              disabled={disabled || input.length === 0}
               onClick={handleOnSubmit}
             >
               <Icon glyph="Plus" />
@@ -77,6 +96,7 @@ const SearchBar: React.FC<SearchBarProps> = ({
             </IconWithTooltip>
           )
         }
+        inputRef={inputRef}
         onChange={(e) => setInput(e.target.value)}
         onKeyPress={(e: KeyboardEvent<HTMLInputElement>) =>
           e.key === "Enter" && handleOnSubmit()
--- a/src/components/TextInputWithGlyph/index.tsx
+++ b/src/components/TextInputWithGlyph/index.tsx
@@ -4,14 +4,15 @@ import { size } from "constants/tokens";
 
 type TextInputWithGlyphProps = {
   icon: React.ReactElement;
+  inputRef?: React.RefObject<HTMLInputElement>;
 } & React.ComponentProps<typeof TextInput>;
 
 const TextInputWithGlyph: React.FC<TextInputWithGlyphProps> = (props) => {
-  const { icon, ...rest } = props;
+  const { icon, inputRef, ...rest } = props;
 
   return (
     <TextInputWrapper>
-      <TextInput {...rest} />
+      <TextInput ref={inputRef} {...rest} />
       <IconWrapper>{icon}</IconWrapper>
     </TextInputWrapper>
   );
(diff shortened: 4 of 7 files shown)


src/components/FiltersDrawer/index.tsx
```diff
@@ -23,6 +24,8 @@ const FiltersDrawer: React.FC<FiltersDrawerProps> = ({ "data-cy": dataCy }) => {
     Cookie.get(HAS_OPENED_DRAWER) === "true"
   );
 
+  useKeyboardShortcut(["Control", "d"], () => setCollapsed(!collapsed));
```
khelif96, line 27: The FilterDrawer already has a keyboard shortcut by default the `[` key it may be confusing to add 2 different shortcuts.

src/components/SearchBar/SearchBar.test.tsx
```diff
@@ -122,4 +122,14 @@ describe("searchbar", () => {
     expect(onChange).toHaveBeenCalledTimes(1);
     expect(onChange).toHaveBeenCalledWith("search", "test1");
   });
+  it("pressing Control+F puts focus on the input", async () => {
```
khelif96, line 125: Possibly another test for cmd + f for brevity.

src/components/TextInputWithGlyph/index.tsx
```diff
@@ -4,14 +4,15 @@ import { size } from "constants/tokens";
 
 type TextInputWithGlyphProps = {
   icon: React.ReactElement;
+  inputRef?: React.RefObject<HTMLInputElement>;
 } & React.ComponentProps<typeof TextInput>;
 
 const TextInputWithGlyph: React.FC<TextInputWithGlyphProps> = (props) => {
```
khelif96, line 10: The right way is to use forwardRef when passing refs into functional components. 
image

src/hooks/useKeyboardShortcut/index.ts
```diff
@@ -0,0 +1,74 @@
...
+type UseKeyboardShortcutOptions = {
+  preventDefault?: boolean;
+  overrideIgnore?: boolean;
+};
+
+// Used to prevent shortcuts from being activated on input elements.
+const INPUT_ELEMENTS = ["INPUT", "TEXTAREA", "SELECT"];
+
+const useKeyboardShortcut = (
+  shortcutKeys: string[],
```
khelif96, line 12: Would love if we could get some type safety here. 
```suggestion
  shortcutKeys: (SpecialKeys | Char)[],
```
Where SpecialKeys would be `type SpecialKeys = "Control" | "Alt" | "Shift";`
And Char would be `type Char = "a" | "b" ....`
Would be a lil verbose but would give us typesafety.

src/hooks/useKeyboardShortcut/useKeyboardShortcut.test.tsx
```diff
@@ -0,0 +1,74 @@
+import { renderHook } from "@testing-library/react-hooks";
+import { render, screen, userEvent } from "test_utils";
+import useKeyboardShortcut from ".";
+
+describe("useKeyboardShortcut", () => {
```
khelif96, line 5: Great tests!

src/components/SearchBar/index.tsx
```diff
@@ -31,14 +32,32 @@ const SearchBar: React.FC<SearchBarProps> = ({
...
+    () => {
+      if (inputRef.current) inputRef.current.focus();
+    },
+    disabled
+  );
+
+  useKeyboardShortcut(
+    ["Control", "s"],
+    () =>
+      selected === "search"
```
khelif96, line 49:
```suggestion
      selected === SearchBarActions.Search
```

src/hooks/useKeyboardShortcut/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  cb: () => void,
+  disabled: boolean = false,
```
khelif96, line 14: I think it would make sense for disabled to be a part of options.

src/hooks/useKeyboardShortcut/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      ) {
+        return false;
+      }
+      if (shortcutKeys.includes("Alt") && !event.altKey) return false;
+      if (shortcutKeys.includes("Shift") && !event.shiftKey) return false;
+
+      return shortcutKeys[shortcutKeys.length - 1] === event.key;
+    },
+    [shortcutKeys]
+  );
```
khelif96, line 36: The behavior of this isn't consistent if a user doesn't pass in a special key.  For example if i want to trigger this hook off of only one key press i can no longer use an input with that character.

In the below example i set a keyboard shortcut for the "s" and "f" keys i can no longer use the text input and type those characters.

https://user-images.githubusercontent.com/4605522/195444299-e3ad79b7-3cd9-46ef-859a-a1336436f8c2.mov

An example when this may come up if we introduce a keyboard shortcut to allow a user to paginate  across their search results e.g. `press "n" to paginate forward or "p" for back`

khelif96 (changes requested):

src/components/FiltersDrawer/index.tsx
minnakt: oops sorry I didn't know that

src/hooks/useKeyboardShortcut/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      if (shouldControl && !event.ctrlKey && !event.metaKey) return false;
+      if (!shouldControl && (event.ctrlKey || event.metaKey)) return false;
+
+      const shouldAlt = shortcutKeys.includes(ModifierKey.Alt);
+      if (shouldAlt && !event.altKey) return false;
+      if (!shouldAlt && event.altKey) return false;
+
+      const shouldShift = shortcutKeys.includes(ModifierKey.Shift);
+      if (shouldShift && !event.shiftKey) return false;
+      if (!shouldShift && event.shiftKey) return false;
```
minnakt, line 50: I don't know how to make this method less ugly 😔

The shortcut keys pressed need to be exact, i.e. pressing `Ctrl + Shift + F` should not activate the shortcut for `Ctrl + F`. This is kind of why I preferred the strategy of keeping track of `heldKeys[]`, but as I explained before that will not work

src/hooks/useKeyboardShortcut/index.ts
```diff
@@ -0,0 +1,91 @@
...
+type UseKeyboardShortcutOptions = {
+  disabled?: boolean;
+  preventDefault?: boolean;
+  overrideIgnore?: boolean;
+};
+
+// Used to prevent shortcuts from being activated on input elements.
+const INPUT_ELEMENTS = ["INPUT", "TEXTAREA", "SELECT"];
+
+const useKeyboardShortcut = (
```
khelif96, line 13: I played around with this a bit what do you think about this?

I separated out the special modifier keys since those have their own fields in an event. It also makes it easy to check for them. This ensures we only run if every modifier key is pressed but we also limit it to only one character. 

```ts

const useKeyboardShortcut = (
  keys: Keys,
  cb: () => void,
  options: UseKeyboardShortcutOptions = {}
) => {
  if (!keys.modifierKeys?.length && !keys.charKey?.length) {
    throw new Error("Must provide at least one key");
  }
  const {
    disabled = false,
    preventDefault = true,
    overrideIgnore = false,
  } = options;
  const cbRef = useRef(cb);
  cbRef.current = cb;

  const handleKeydown = useCallback(
    (event: KeyboardEvent) => {
      let areModifierKeysPressed = true;
      if (keys.modifierKeys && keys.modifierKeys.length) {
        areModifierKeysPressed = getPressedModifierKeys(event).every(
          (modifierKey) => keys.modifierKeys?.includes(modifierKey)
        );
      }
      const isCharKeyPressed = keys.charKey === event.key;

      const shouldExecute =
        overrideIgnore ||
        !INPUT_ELEMENTS.includes((event.target as HTMLElement).tagName);

      console.log(areModifierKeysPressed, isCharKeyPressed, shouldExecute);
      if (areModifierKeysPressed && isCharKeyPressed) {
        if (shouldExecute) {
          // Prevent browser default behavior.
          if (preventDefault) {
            event.preventDefault();
          }
          cbRef.current();
        }
      }
    },
    [keys, preventDefault, overrideIgnore]
  );

  useEffect(() => {
    // There's no need to keep track of events if the component is disabled.
    if (disabled) {
      document.removeEventListener("keydown", handleKeydown);
    } else {
      document.addEventListener("keydown", handleKeydown);
    }
    return () => {
      document.removeEventListener("keydown", handleKeydown);
    };
  }, []);
};

// Function checks which modifier keys are pressed and returns an array of the pressed modifier keys
const getPressedModifierKeys = (event: KeyboardEvent) => {
  const pressedModifierKeys: ModifierKey[] = [];
  if (event.ctrlKey) {
    pressedModifierKeys.push(ModifierKey.Control);
  }
  if (event.altKey) {
    pressedModifierKeys.push(ModifierKey.Alt);
  }
  if (event.shiftKey) {
    pressedModifierKeys.push(ModifierKey.Shift);
  }
  return pressedModifierKeys;
};
```

khelif96 (changes requested):
What do you think about my suggestion below it cleans up some of the code yo uwrote and handles your scenario.

src/constants/keys.ts
```diff
@@ -0,0 +1,43 @@
...
+  Q = "q",
+  R = "r",
+  S = "s",
+  T = "t",
+  U = "u",
+  V = "v",
+  X = "x",
+  Y = "y",
+  Z = "z",
+
```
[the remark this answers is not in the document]
khelif96: I Specifically meant also including the `?` character since we will use that for the keyboard shortcut tooltip.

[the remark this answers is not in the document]
minnakt: Correct me if I'm wrong, but I think in that case we would register the `Shift` + `Forward Slash`

khelif96 (approved):

PR #70, opened 2022-10-11
EVG-17538: Make it possible to expand collapsed rows

minnakt:
EVG-17538

### Description 
This PR makes it possible to expand collapsed rows.

Notable changes:
* Moved `CollapsedRow` out of `AnsiiRow` and `ResmokeRow`, it now stands alone.
* You will need to toggle on the Expandable Rows mode in order to see collapsed rows.

### Screenshots
Just for the sake of showing how re-collapsing rows could work, I listed the expanded ranges in the Filters Drawer. I will delete this code before merging the PR.

Screen Shot 2022-10-11 at 2 22 08 PM

### Testing 
- Added tests in `ansii_logView.ts`
- Added tests in `resmoke_logView.ts`
- Added tests in `CollapsedRow.test.tsx`
- Separated `CollapsedRow` story into `CollapsedRow.stories.tsx`
- Added tests in `expandedRanges.test.ts`


src/components/FiltersDrawer/index.tsx
```diff
@@ -1,24 +1,32 @@
...
 const { green, gray } = palette;
 
 interface FiltersDrawerProps {
   ["data-cy"]?: string;
+  expandedLines: ExpandedLines;
+  collapseLines: (idx: number) => void;
 }
 
-const FiltersDrawer: React.FC<FiltersDrawerProps> = ({ "data-cy": dataCy }) => {
+const FiltersDrawer: React.FC<FiltersDrawerProps> = ({
```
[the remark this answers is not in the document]
minnakt: I think it's okay because Parsley is on React 18 ^^ Spruce is on 17, so it needs to use `VFC`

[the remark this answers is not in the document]
khelif96: Yup the change was made in the update to react 18. `React.FC` no longer includes children by default and is the recommended way to go. https://github.com/DefinitelyTyped/DefinitelyTyped/commit/55dc209ceb6dbcd59c4c68cc8dfb77faadd9de12#diff-32cfd8cb197872bcba371f5018185d2e75fa540b52cda2dd7d8ac12dcc021299L500

We used React.VFC on Spruce since the typings for React 17 explicitly typed children which is error prone.

src/components/LogRow/CollapsedRow/index.tsx
```diff
@@ -3,47 +3,83 @@ import styled from "@emotion/styled";
...
+      <StyledButton
+        leftGlyph={<Icon glyph="Expand" />}
+        onClick={expandAll}
+        size="xsmall"
+      >
         All
       </StyledButton>
-      <StyledButton leftGlyph={<Icon glyph="Expand" />} size="xsmall">
-        5 Above and Below
+      <StyledButton
```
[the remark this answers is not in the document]
minnakt: oop yes just needs to be added to a dependency array

src/components/Icon/icons/index.tsx
```diff
@@ -8,7 +11,7 @@ interface LeafygreenIconProps extends React.SVGProps<SVGSVGElement> {
 
 export const ClosedEye: React.ComponentType<LeafygreenIconProps> = ({
```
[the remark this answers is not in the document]
minnakt: yes I noticed!!! but I just like Will's cute custom icon so much 😭

sophstad (approved):

src/components/LogRow/CollapsedRow/index.tsx
```diff
@@ -3,47 +3,83 @@ import styled from "@emotion/styled";
...
+  const expandFive = () => {
+    expandLines([
+      [start, start + (SKIP_NUMBER - 1)],
+      [end - (SKIP_NUMBER - 1), end],
+    ]);
+  };
+
+  const expandAll = () => {
+    expandLines([[start, end]]);
+  };
```
khelif96, line 38: We can considerably speed this up by taking advantage of some of the concurrency features built into react 18
https://reactjs.org/docs/hooks-reference.html#usetransition
```suggestion
  const [isPending, startTransition] = useTransition();
  const expandFive = () => {
  startTransition(() => {
    expandLines([
      [start, start + (SKIP_NUMBER - 1)],
      [end - (SKIP_NUMBER - 1), end],
    ]);
    });
  };

  const expandAll = () => {
   startTransition(() => {
    expandLines([[start, end]]);
    });
  };
```

src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -6,17 +6,22 @@ import {
...
 
 const RowRenderer: RowRendererFunction = (data) => {
-  const { logType } = data;
-  const Row = rowRendererMap[logType];
+  const { logType, processedLines } = data;
+
   const result = (props: ListRowProps) => {
     const { index, key, parent } = props;
 
+    const Row = Array.isArray(processedLines[index])
```
khelif96, line 21: `isCollapsedRow`

src/context/LogContext/index.tsx
```diff
@@ -82,15 +89,29 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
+    [
+      state.logs.length,
+      `${filters}`, // eslint-disable-line react-hooks/exhaustive-deps
+      `${bookmarks}`, // eslint-disable-line react-hooks/exhaustive-deps
+      `${state.expandedLines}`, // eslint-disable-line react-hooks/exhaustive-deps
+      selectedLine,
+      filterLogic,
+      expandableRows,
+    ]
   );
```
khelif96, line 112: We can use the new [`useDeferredValue`](https://reactjs.org/docs/hooks-reference.html#usedeferredvalue) hook to defer this state update. This function is the slowest part of the context and is what causes the app to lag when rendering it. We are already memoizing its value to prevent it from rerendering but we can also defer its updates when it does run to avoid blocking the UI. In my experience this lead to a pretty dramatic improvement in performance. 
```ts
  const deferredFilters = useDeferredValue(filters);
  
  // TODO EVG-17537: more advanced filtering
  const processedLogLines = useMemo(
    () => {
      return filterLogs({
        logLines: state.logs,
        filters,
        bookmarks,
        selectedLine,
        filterLogic,
        expandedLines: state.expandedLines,
        expandableRows,
      });
    },
    [
      state.logs.length,
      `${deferredFilters}`, // eslint-disable-line react-hooks/exhaustive-deps
      `${bookmarks}`, // eslint-disable-line react-hooks/exhaustive-deps
      `${state.expandedLines}`, // eslint-disable-line react-hooks/exhaustive-deps
      selectedLine,
      filterLogic,
      expandableRows,
    ]
  );



```

src/types/logs.ts
```diff
@@ -1,5 +1,6 @@
-type ProcessedLogLine = number | number[];
+type ExpandedLines = Array<[number, number]>;
```
khelif96, line 1: nit. So we can use ExpandedLine in the future if we need it.
```suggestion
type ExpandedLine = [number, number];
type ExpandedLines = ExpandedLine[];
```

src/components/LogRow/ResmokeRow/index.tsx
```diff
@@ -18,13 +16,7 @@ const ResmokeRow = forwardRef<any, BaseRowProps>((rowProps, ref) => {
...
   const { index } = listRowProps;
 
-  const line = processedLines[index];
-
-  if (isCollapsedRow(line)) {
-    return (
-      <CollapsedRow ref={ref} {...listRowProps} numCollapsed={line.length} />
-    );
-  }
+  const line = processedLines[index] as number;
```
khelif96, line 19: Since we do this check inside of RowRenderer now i wonder if we can update the type for this so we don't have to do this type assertion.

src/components/LogRow/AnsiiRow/index.tsx
```diff
@@ -20,15 +19,8 @@ const AnsiiRow = forwardRef<any, BaseRowProps>((rowProps, ref) => {
...
-  const { index } = listRowProps;
-
-  const line = processedLines[index];
 
-  if (isCollapsedRow(line)) {
-    return (
-      <CollapsedRow ref={ref} {...listRowProps} numCollapsed={line.length} />
-    );
-  }
+  const line = processedLines[index] as number;
```
khelif96, line 23: Same comment as below

khelif96 (changes requested):
We should take advantage of some of [React 18's new concurrent features](https://17.reactjs.org/docs/concurrent-mode-intro.html#what-is-concurrent-mode) in my experience trying them out with your changes they have a night and day impact.  There are some quirks with how the list renders but those can likely be solved with a useEffect.


Before:

https://user-images.githubusercontent.com/4605522/196511727-5ce42a27-fb64-4ff2-b249-f0523e79a021.mov



After:

https://user-images.githubusercontent.com/4605522/196511963-70ffcc88-d921-4916-9239-b302480669e9.mov

minnakt (commented):
Reviewing for update

minnakt:
Another update comment:
- I tried it without `useDeferredValue`, I think it works okay? What do you think, can we forgo these? 🤔
- The reason why expanding rows wasn't jumping was because it wasn't in the dependency array to resize rows 😭 However there's an alternate solution: have all the rows be the same size. The collapsed rows look terrible, but hey, it works... 🙃
- I removed `filter.test.ts` because I don't really want to fix the tests for it... I'll be rewriting all of these tests in the next PR.

src/context/LogContext/index.tsx
```diff
@@ -64,9 +81,50 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
+      `${selectedLine}`, // eslint-disable-line react-hooks/exhaustive-deps
+      `${state.expandedLines}`, // eslint-disable-line react-hooks/exhaustive-deps
+      expandableRows,
+    ]
+  );
+
+  useEffect(() => {
+    cache.clearAll();
+    listRef.current?.recomputeRowHeights();
+  }, [wrap, matchingLines, expandableRows]);
```
khelif96, line 126: Can we move this back to LogPane/LogWindow?

src/utils/filter/index.ts
```diff
@@ -1,59 +1,57 @@
...
  */
-export const filterLogs = (
-  logLines: string[],
-  filters: string[],
-  bookmarks: number[],
-  selectedLine: number | undefined,
-  filterLogic: FilterLogic
-): ProcessedLogLines => {
-  if (filters.length === 0) {
+export const filterLogs = ({
```
khelif96, line 26: Tests please!

src/utils/filter/index.ts
```diff
@@ -70,3 +68,35 @@ export const filterLogs = (
...
 };
+
+/**
+ * Function that determines if a particular log line satisfies the filter conditions.
+ * @param logLines - list of strings representing the log lines
+ * @param filters - list of filters being applied
+ * @param filterLogic - specifies whether to use AND or OR filtering
+ * @returns true if filter conditions are satisfied, false otherwise
+ */
+export const getMatchingLines = (
```
khelif96, line 79: Tests!

khelif96 (approved):
LGTM 🎉  🍰

PR #73, opened 2022-10-17
EVG-17537: Implement advanced filtering

minnakt:
EVG-17537

### Description 
This PR adds advanced filtering which includes toggling visibility, toggling case sensitivity, and toggling match type (exact vs inverse). The bulk of this PR is tests.

#### Explanation of query params 
I took a look at how Lobster constructs the query params for filters and thought it made sense. Visibility, case sensitivity, and match type all have exactly two states so they can be mapped to 0 or 1. Every filter is prefixed by a `0/1` string of length 3, which represents the visibility, case sensitivity, and match type state.

The `0/1` representation helps keep the URL string concise and it's pretty readable. As a reminder, the maximum possible length for a URL is about 2000 characters.

#### Other things
- No more fancy regex. JavaScript does not support partial case sensitivity in regex so I don't even think it's possible. 🤔 I think it's easier to understand without it anyways!
- If there is a different preferred way to construct the query params, I can switch it to that. The underlying logic should stay the same.

### Screenshots

https://user-images.githubusercontent.com/47064971/197250569-165b0ea5-2741-4734-bd42-08eeeaab1368.mov

### Testing 
- Cypress
  - Extracted filtering tests to `ansii_filtering.ts` and `resmoke_filtering.ts`
- Jest
  - Added tests in `matchingLines.test.ts` 
  - Added tests in `useQueryParams.test.ts`
  - Added tests in `query-string.test.ts`
  - Modified some other random Jest tests just to improve code coverage


src/context/toast/toast.test.tsx
```diff
@@ -206,6 +206,31 @@ describe("toast", () => {
...
+      });
+      expect(screen.getByDataCy("toast")).toBeInTheDocument();
+
+      // Advance timer, which would normally trigger timeout.
+      act(() => {
+        jest.advanceTimersByTime(TOAST_TIMEOUT);
+      });
+      await waitFor(() => {
+        expect(screen.getByDataCy("toast")).toBeInTheDocument();
+      });
```
khelif96, line 228: I'm wondering if this works without mocking the timers?

src/hooks/useQueryParam/index.ts
```diff
@@ -36,15 +44,39 @@ const useQueryParam = <T>(
...
+      setSearchParams({
+        ...searchParams,
+        filters: stringifyFilters(filters),
+      });
+    },
+    [setSearchParams, searchParams]
+  );
+
+  return [parsedFilters, setFiltersParam] as const;
 };
```
khelif96, line 80: I think this makes sense to move to its own folder/file. useQueryParm is generic this seems more geared to a specific feature

src/utils/query-string/index.ts
```diff
@@ -8,4 +10,39 @@ export const parseQueryString = (search: string) =>
...
+};
+
+export const stringifyFilters = (filters: Filters): string[] =>
+  filters.map((f) => {
+    const visible = f.visible ? 1 : 0;
+    const caseSensitive = f.caseSensitive === CaseSensitivity.Sensitive ? 1 : 0;
+    const matchType = f.matchType === MatchType.Inverse ? 1 : 0;
+
+    return `${visible}${caseSensitive}${matchType}${f.name}`;
+  });
```
khelif96, line 48: Nice!

src/utils/matchingLines/index.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @returns a set of numbers representing the matched lines, or undefined
+ */
+export const getMatchingLines = (
+  logLines: string[],
+  filters: Filters,
+  filterLogic: FilterLogic
+) => {
+  const set = new Set<number>();
+
+  const visibleFilters = filters.filter((f) => f.visible !== false);
```
khelif96, line 68: It doesn't look like the filter rules are being obeyed. For example i have one filter that is hidden and one that is visible and i would expect to see the rules from the visible one but instead it seems to be showing no lines.
https://user-images.githubusercontent.com/4605522/197627122-32480cb4-1633-4d2e-8a38-4fba8d5eb89e.mov

khelif96 (changes requested):

src/context/toast/toast.test.tsx
minnakt: Hmm it does, but test is supposed to test that the toast does not disappear even when the `setTimeout` time elapses. I tried doing it a different way by mocking `setTimeout` but that doesn't really work because I think the `useEffect` in toast actually gets called multiple times. In the end I just deleted the test

src/utils/matchingLines/index.ts
minnakt: I think you may have attached the wrong video haha! But yes I was passing the wrong thing down to other functions. I added a test for this 🙏

khelif96 (approved):
LGTM Great job! 🐎

PR #75, opened 2022-10-18
EVG-17545 Create ErrorBoundary and integrate Bugsnag

khelif96:
EVG-17545

### Description 
* Installs Bugsnag and creates an error boundary .
* Runs Bugsnag on all deployable environments
* Adds a `-custom` flag to builds so we can identify if a bugsnag build reflects something that is in main or if its a custom build.
* Left a few breadcrumbs so we can more easily trace and diagnose failures.


minnakt (commented):
Do you want to add the tests for `ErrorBoundary` and `errorReporting` from Spruce as well? also i hope you dont mind me updating your PR title 🙏😇

src/components/ErrorBoundary/ErrorBoundary.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+  component: ErrorFallback,
+};
+
+export const Template: ComponentStory<typeof ErrorFallback> = () => (
+  <ErrorFallback />
+);
+
+Template.storyName = "Error Fallback";
+
+export const ErrorBoundaryTemplate: ComponentStory<
```
minnakt, line 17: I'm not sure if it's worth having this story, just because I think you'll need to add an extra script in `evergreen.yml` to upload the SVG for it to even appear on Storybook Production

minnakt (approved):
lgtmeeeeeee

src/components/ErrorBoundary/ErrorBoundary.stories.tsx
khelif96: It's ok I can add the script with some copy+pasta

PR #76, opened 2022-10-18
EVG-18081 Sanitize html characters for resmoke lines

khelif96:
EVG-18081

### Description 
Some resmoke lines often contain strings that resemble html characters. This is problematic when they are run through `parseHtml` since it will try to transform the string into an html element which does not exist. This PR adds a function called `escapeHtml`  which will replace any special characters with the ansii string representation. 

Added an allowlist to `parseHtml` i only enabled the tags we would expect to see this can be updated in the future.

### Testing 
No more console errors
Unit tests

### DIFF
--- a/src/components/LogRow/ResmokeRow/index.tsx
+++ b/src/components/LogRow/ResmokeRow/index.tsx
@@ -2,6 +2,7 @@ import { forwardRef } from "react";
 import CollapsedRow from "components/LogRow//CollapsedRow";
 import BaseRow from "components/LogRow/BaseRow";
 import { isCollapsedRow } from "utils/collapsedRow";
+import { escapeHtml } from "utils/renderHtml/escapeHtml";
 import { BaseRowProps } from "../types";
 import { isLineInRange } from "../utils";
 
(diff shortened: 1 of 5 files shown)


src/components/LogRow/ResmokeRow/index.tsx
```diff
@@ -26,7 +27,7 @@ const ResmokeRow = forwardRef<any, BaseRowProps>((rowProps, ref) => {
       <CollapsedRow ref={ref} {...listRowProps} numCollapsed={line.length} />
     );
   }
-  const lineContent = getLine(line);
+  const lineContent = escapeHtml(getLine(line) || "");
```
minnakt, line 30: I wonder if we should do this when processing the resmoke lines in `LogContext/state.ts`. What do you think? 🤔

src/utils/renderHtml/escapeHtml.ts
```diff
@@ -0,0 +1,9 @@
+const escapeHtml = (html: string) =>
```
minnakt, line 1: sorry one more thing; I noticed the highlight doesn't appear when searching for things with escaped characters, though the search picks it up correctly

Screen Shot 2022-10-19 at 10 11 53 AM

src/components/LogRow/ResmokeRow/index.tsx
khelif96: I moved it to baseRow instead. All the characters are technical real characters so we need to be able to search them and replacing them in `LogContext/state.ts` will not allow us to do that anymore.

src/utils/renderHtml/escapeHtml.ts
khelif96: Yup i ended up moving the escape to after we find the match and wrap it in `<mark>`

minnakt (approved):
yeehaw 🤠

PR #77, opened 2022-10-21
EVG-18010 Add log highlighting support

khelif96:
EVG-18010

### Description 
Adds a new option to searchbar to apply a highlight. 
Add highlights to filter drawer.
Uses same style as search highlight

### Screenshots
image


src/components/FiltersDrawer/index.tsx
```diff
@@ -130,12 +139,48 @@ const FiltersDrawer: React.FC<FiltersDrawerProps> = ({
...
               <Body>No lines have been expanded.</Body>
-            </ExpandedLineWrapper>
+            </SectionWrapper>
+          )}
+        </StyledSideNavGroup>
+
+        <StyledSideNavGroup
+          glyph={
+            <Icon
+              fill={highlights.length ? green.dark2 : gray.base}
```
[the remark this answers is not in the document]
khelif96: Great idea! Although i had to update its displayName property because leafygreen doesn't like glyph not being an Icon

src/components/FiltersDrawer/index.tsx
```diff
@@ -130,12 +139,48 @@ const FiltersDrawer: React.FC<FiltersDrawerProps> = ({
...
+              glyph="Minus"
```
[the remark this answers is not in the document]
khelif96: Yeah i wasn't sure what to put i can bring it up in the design sync

cypress/integration/ansiiLogs/ansii_logView.ts
```diff
@@ -284,6 +284,59 @@ describe("Searching", () => {
   });
 });
 
+describe("Highlighting", () => {
```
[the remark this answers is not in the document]
khelif96: The question is "who will merge there's in FIRST?" .... jk sure i'll move it over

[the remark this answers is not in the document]
minnakt: i need an angry emote on github

src/components/FiltersDrawer/index.tsx
```diff
@@ -130,18 +131,58 @@ const FiltersDrawer: React.FC<FiltersDrawerProps> = ({
...
           ) : (
-            <ExpandedLineWrapper data-cy="no-expanded-lines-message">
+            <SectionWrapper data-cy="no-expanded-lines-message">
               <Body>No lines have been expanded.</Body>
-            </ExpandedLineWrapper>
+            </SectionWrapper>
+          )}
+        </StyledSideNavGroup>
+
+        <StyledSideNavGroup
```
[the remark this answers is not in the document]
khelif96: We discussed this offline but i agree it would be really nice to refactor this out. The LG SideNavGroup component is being pretty difficult so i am going to defer this to a follow up ticket.

sophstad (approved):

src/components/LogRow/BaseRow/index.tsx
```diff
@@ -128,13 +133,26 @@ const ProcessedBaseRow: React.FC<ProcessedBaseRowProps> = memo((props) => {
...
+      const shouldCheckForOverlappingRegex = searchTerm !== undefined;
+      const hasOverlappingRegexes =
+        shouldCheckForOverlappingRegex &&
+        hasOverlappingRegex(searchTerm, highlights, children);
+
+      if (!hasOverlappingRegexes) {
+        render = render.replace(
+          highlights,
+          (match) => `<mark>${escapeHtml(match)}</mark>`
+        );
```
minnakt, line 146: I think there might be a small bug with the `replace` call, because it only highlights the first instance of the word rather than all of them. In the screenshot below, I've added `uuid` as a highlight but only the first instance in each line gets highlighted. 

Screen Shot 2022-10-28 at 11 29 35 AM


This differs from Lobster's behavior, which highlights all instances of the highlight word in each line (the highlight word is `project` here):

Screen Shot 2022-10-28 at 11 28 54 AM

I think this also holds true for search terms.

src/utils/regex/index.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * hasOverlappingRegex - Checks if 2 regexes overlap each other
```
minnakt, line 2: nit. for consistency
```suggestion
 * `hasOverlappingRegex` - Checks if 2 regexes overlap each other
```

src/components/FiltersDrawer/index.tsx
```diff
@@ -141,18 +144,58 @@ const FiltersDrawer: React.FC<FiltersDrawerProps> = ({
...
+                    aria-label="Delete highlight"
+                    data-cy="delete-highlight-button"
+                    onClick={() => deleteHighlight(highlight)}
+                  >
+                    <Icon fill={green.dark2} glyph="X" />
+                  </IconButton>
+                  <Overline>{highlight}</Overline>
+                </NonFilterElement>
+              ))}
+            </>
```
minnakt, line 179: nit. unnecessary fragment (lint is probably very wrong sry)
```suggestion
highlights.map((highlight) => (
              <NonFilterElement key={`highlight-${highlight}`}>
                <IconButton
                  aria-label="Delete highlight"
                  data-cy="delete-highlight-button"
                  onClick={() => deleteHighlight(highlight)}
                >
                  <Icon fill={green.dark2} glyph="X" />
                </IconButton>
                <Overline>{highlight}</Overline>
              </NonFilterElement>
            ))
```

minnakt (changes requested):
Ready to approve once bug is addressed!

I think it's worth mentioning that on legacy Lobster, highlights can be toggled on or off, be case sensitive or insensitive, and be toggled for word or line. However I'm kind of okay omitting these features because we don't have designs, and I'm wondering who even uses them anyways 😒 If no one complains about it during user testing, then I say good riddance

src/components/LogRow/BaseRow/index.tsx
khelif96: Good catch

khelif96: Great catch i fixed this!

src/components/LogRow/BaseRow/index.tsx
```diff
@@ -114,27 +118,40 @@ interface ProcessedBaseRowProps {
...
+      const hasOverlappingRegexes =
+        shouldCheckForOverlappingRegex &&
+        hasOverlappingRegex(searchTerm, highlights, children);
+      if (!hasOverlappingRegexes) {
+        render = render.replaceAll(
+          new RegExp(highlights, "g"),
+          (match) => `<mark>${escapeHtml(match)}</mark>`
+        );
+      }
+    }
```
minnakt, line 147: this is possibly a really dumb comment lol... but I just wanted to leave it as food for thought. I was just trying to see if we could reduce some of the complicated logic. This passes all of the tests written for `BaseRow`.
```suggestion
    // I stole this expression from spruce
    const impossibleRegex = /^\b$/;

    const searchTermRegex = searchTerm
      ? new RegExp(searchTerm)
      : impossibleRegex;

    const highlightRegex = highlights
      ? new RegExp(highlights)
      : impossibleRegex;

    // Add "i" if case is insensitive. I am assuming the case applies to both search and highlights since highlights can no longer manage their own case sensitivity. However if this assumption doesn't hold then this solution is meaningless.
    const replaceRegex = new RegExp(
      `${searchTermRegex.source}|${highlightRegex.source}`,
      "gi"
    );

    const render = children.replace(
      replaceRegex,
      (match) => `<mark>${escapeHtml(match)}</mark>`
    );
```

src/utils/regex/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param regex2
+ * @param text
+ * @returns true if regexes overlap, false otherwise
+ */
+const hasOverlappingRegex = (regex1: RegExp, regex2: RegExp, text: string) => {
+  const regex1Matches = text.match(regex1);
+  const regex2Matches = text.match(regex2);
+  if (regex1Matches && regex2Matches) {
+    return stringIntersection(regex1Matches[0], regex2Matches[0]);
+  }
```
minnakt, line 15: can you add a test that specifically addresses this change? i.e. imagine that I'm dumb in the future and I think that I can just remove it because the tests don't fail

src/components/LogRow/BaseRow/index.tsx
```diff
@@ -114,27 +118,40 @@ interface ProcessedBaseRowProps {
...
-  const { children, searchTerm, color, "data-cy": dataCy } = props;
+  const { children, searchTerm, color, "data-cy": dataCy, highlights } = props;
   const memoizedLogLine = useMemo(() => {
     let render = children;
     if (searchTerm) {
       // escape the matching string to prevent XSS
-      render = render.replace(
-        searchTerm,
+      render = render.replaceAll(
+        new RegExp(searchTerm, "g"),
```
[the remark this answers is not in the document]
khelif96: So `replaceAll` requires the "g" flag when passing in a regex, but replace with the "g" flag works the same way.  From what I can gather the only difference is that `replaceAll` will also do a global replacement if you pass a string without regex.

src/components/LogRow/BaseRow/index.tsx
khelif96: Good catch i fixed the sensitivity bug!

src/utils/regex/index.ts
khelif96: Sure just added a test to regex.test.ts

PR #79, opened 2022-10-24
EVG-17682: Add pretty print bookmarks functionality

minnakt:
EVG-17682

### Description 
This PR allows bookmarked lines to be pretty printed. Initially I wanted to use the [`mongodb/js-bson`](https://github.com/mongodb/js-bson) library to do this, but as it turns out, it does not support preserving large numbers in JavaScript. ☹️ Instead I used [`lossless-json`](https://github.com/josdejong/lossless-json) to parse the lines. This is slower than the native `JSON.parse` and `JSON.stringify` functions but the number of un-bookmarked lines should far outweigh the number of bookmarked lines.

This PR makes use of clearing the cache at a specific index instead of clearing the whole cache. If we didn't do this, the scroll would be jumping around every time someone made a pretty print bookmark which would make the application borderline unusable.

I also re-wrote some of the styling of `BaseRow`. Turning on wrap should now prevent the line from falling under the line number text.

### Screenshots
Screen Shot 2022-10-24 at 9 55 24 AM

### Testing 
- Added a test for pretty printing in `resmoke_logView.ts`
- Added tests for expanded rows that were missing (Cypress)


src/components/LogRow/BaseRow/index.tsx
```diff
@@ -99,14 +103,16 @@ const BaseRow = forwardRef<any, BaseRowProps>((props, ref) => {
...
-        {children}
-      </ProcessedBaseRow>
-    </StyledPre>
+      <StyledPre shouldWrap={wrap}>
+        <ProcessedBaseRow
+          color={resmokeRowColor}
+          data-cy={dataCyText}
+          searchTerm={searchTerm}
+        >
+          {bookmarked && prettyPrint ? formatPrettyPrint(children) : children}
```
khelif96, line 112: This feature only really makes sense for resmoke rows. I wonder if we should gate it and disable the option for non resmoke rows?

khelif96: If you agree then i think we can move this up a level to ResmokeRow and remove the need for BaseRow to have to know about prettyPrint

src/utils/prettyPrint/index.ts
```diff
@@ -0,0 +1,89 @@
+import { parse, stringify } from "lossless-json";
+
+/**
+ * `findJsonObjects` finds the indices of json objects in a given log line.
+ * @param logLine - the log line in which to search for json objects
+ * @returns an array of [startIndex, endIndex] representing the position of each json object in the string
+ */
+export const findJsonObjects = (logLine: string) => {
```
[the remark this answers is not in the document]
minnakt: I want to say it's slightly different because it seems like `getJsonString` can only get one JSON object in a string. However it is possible for there to be multiple JSON objects in a resmoke line, like below:

Screen Shot 2022-10-25 at 1 01 39 PM

src/components/LogRow/BaseRow/index.tsx
minnakt: So technically I did prevent it from working on `AnsiiRow`s, but I didn't pull it up to `ResmokeRow`... it's because I would also need to pull up the `bookmarks` and `resetRowHeightAtIndex`, which might be a little awkward.

src/utils/prettyPrint/index.ts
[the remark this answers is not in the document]
khelif96: Do you think it would make sense to only use your function or is there still value in having both? If the latter can we add some documentation to both of these on when to use each one?

src/components/LogRow/BaseRow/index.tsx
khelif96: Ok cool can we cut a ticket to investigate refactoring this and other resmoke functionality and put them in PM-3000?

src/utils/prettyPrint/index.ts
```diff
@@ -0,0 +1,89 @@
...
+      logLine.substring(jsonIndices[i][0], jsonIndices[i][1] + 1)
+    ).replace(/"([^"]+)"\s*:/g, "$1:");
+    formatted += "\n";
+
+    currIndex = jsonIndices[i][1] + 1;
+  }
+  if (currIndex !== logLine.length) {
+    formatted += logLine.substring(currIndex);
+  }
+
```
khelif96, line 87: Now that we know we can only have 1 json object does this need to change or does this logic apply to nested json objects?

src/utils/prettyPrint/prettyPrint.test.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    const openBracket = json.indexOf("{");
+    const closedBracket = json.lastIndexOf("}");
+
+    expect(result).toStrictEqual([[openBracket, closedBracket]]);
+  });
+
+  it("should correctly determine json indices when there are multiple json objects", () => {
+    const logLine =
+      '[js_test:backup_restore_rolling] 2020-03-02T08:52:04.781+0000 d20521| {"t":{"$date":"2020-03-02T08:52:04.780+0000"},"s":"I", "c":"RECOVERY","id":23987,"ctx":"initandlisten","msg":"WiredTiger recoveryTimestamp. Ts: {recoveryTimestamp}","attr":{"recoveryTimestamp":{"$timestamp":{"t":0,"i":0}}}}and then some more text{"std_get_0_envDataEntry":"distmod","std_get_1_envDataEntry":"rhel62"}';
```
khelif96, line 24: Lets omit this now that we can only have one json object in a resmoke line

khelif96 (approved):
Great work just had some comments about the json parsing now that we know we can only have 1 json object

src/components/LogRow/BaseRow/index.tsx
minnakt: added [EVG-18167](https://jira.mongodb.org/browse/EVG-18167) 💃

PR #80, opened 2022-10-24
EVG-18070 Fix memory leak when downloading logs using axios

khelif96:
EVG-18070

### Description 
For some reason downloading logs with Axios would cause the log file to be kept in memory twice even after the download components get unmounted. 

Memory usage and download time downloading a 237 mb resmoke log file
Before:
Download Time 2.42
Heap snapshot size: 465 mb
image

After:
Download Time 2.41 s
Heap snapshot size: 231 mb
image

You may say "Woah! it's smaller than the actual log file we downloaded" That's likely due to the processing we perform on the file we strip out some parts of the log that aren't important.


src/hooks/useLogDownloader/useLogDownloader.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+// Fetch is not supported in jest so we need to mock it out
+describe("useLogDownloader", () => {
+  it("gets a good response from the api and updates its state", async () => {
+    const mockFetchPromise = Promise.resolve({
+      ok: true,
+      text: () => Promise.resolve(textMessage),
+    });
+    // eslint-disable-next-line jest/prefer-spy-on -- we want to mock the global fetch and we can't spy on it
+    global.fetch = jest.fn().mockImplementation(() => mockFetchPromise);
```
minnakt, line 17: Does this work?

```suggestion
    const mockFetchPromise = jest.fn().mockResolvedValue({
      ok: true,
      text: () => Promise.resolve(textMessage),
    });

    jest.spyOn(global, "fetch").mockImplementation(mockFetchPromise);
```

src/hooks/useLogDownloader/useLogDownloader.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    expect(result.current.data).toStrictEqual([
+      "Fetched a multiline log file",
+      "Some more lines",
+      "And some more",
+    ]);
+  });
+  it("gets a bad response from the api and returns an error", async () => {
+    const mockFetchPromise = Promise.reject(new Error("Something went wrong"));
+    // eslint-disable-next-line jest/prefer-spy-on -- we want to mock the global fetch and we can't spy on it
+    global.fetch = jest.fn().mockImplementation(() => mockFetchPromise);
```
minnakt, line 34: Does this work?

```suggestion
    const mockFetchPromise = jest
      .fn()
      .mockRejectedValue(new Error("Something went wrong"));

    jest.spyOn(global, "fetch").mockImplementation(mockFetchPromise);
```

jest.config.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
     "^uuid$": "<rootDir>/node_modules/uuid/dist/index.js",
   },
   modulePaths: ["<rootDir>/src"],
+  setupFiles: ["<rootDir>/setupJest.js"],
```
minnakt, line 13: I'm not a huge fan of these one line files 😓 Can we do this instead and delete the file?
```suggestion
  setupFiles: ["whatwg-fetch"],
```

src/pages/LogView/LoadingPage/index.tsx
```diff
@@ -139,9 +142,15 @@ const LoadingBarContainer = styled.div`
...
-  margin-bottom: ${size.s};
 `;
 
+const SpaceBetween = styled.div`
+  margin-bottom: ${size.s};
+  display: flex;
+  justify-content: space-between;
+  width: 100%;
+  align-items: flex-end;
+`;
```
minnakt, line 153: I don't think we need this styling; I think you can just slap a `gap` property on `LoadingBarContainer` to achieve the spacing

src/hooks/useLogDownloader/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    })
+      .then((response) => {
+        if (!response.ok) {
+          throw new Error(response.statusText);
+        }
+        return response;
+      })
+      .then((response) => response.text() || "")
+      .then((text) => {
+        setData(text.split("\n"));
```
minnakt, line 39: I think the tests are failing because we don't call `trimEnd`... I guess an alternative would be to decrement `lineCount` by 1 in anticipation of the new line at the end, though that wouldn't work if we were to chunk logs in the future. 😔

minnakt (changes requested):

jest.config.js
khelif96: Good one!

src/hooks/useLogDownloader/useLogDownloader.test.ts
khelif96: Good catch! this didn't work before but i guess that was before i install whatwg-fetch

src/pages/LogView/LoadingPage/index.tsx
khelif96: Nice one! CSS PRO

src/hooks/useLogDownloader/index.ts
```diff
@@ -0,0 +1,56 @@
...
+        setError(err.message);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+    return () => {
+      abortController.abort();
+    };
+  }, [url]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
minnakt, line 52: Seems like you can delete this!

```suggestion

```

src/hooks/useLogDownloader/index.ts
```diff
@@ -0,0 +1,56 @@
+import { useEffect, useState } from "react";
+import { isProduction } from "utils/environmentVariables";
+import { leaveBreadcrumb } from "utils/errorReporting";
+/*
```
minnakt, line 4: nit. missing a star here, causing this not to be recognized as a JSDoc comment
```suggestion
/**
```

src/hooks/useLogDownloader/index.ts
```diff
@@ -0,0 +1,56 @@
...
+ * useLogDownloader is a custom hook that downloads a log file from a given URL.
```
minnakt, line 5: nit. for consistency
```suggestion
 * `useLogDownloader` is a custom hook that downloads a log file from a given URL.
```

src/hooks/useLogDownloader/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  const [data, setData] = useState<string[] | null>(null);
+  const [error, setError] = useState<string | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    leaveBreadcrumb("useLogDownloader", { url }, "request");
+    const req = new Request(url, { method: "GET" });
+    // We can utilize AbortController to cancel the request if the component unmounts
+    // When running in strict mode this will trigger an error because of how strict mode behaves
+    // We need to case on it to avoid the error
```
minnakt, line 24: Let's move these comments closer to where they apply. So for example, I think you can do:
* put comment `"Cancel the request if the component unmounts"` above` L48`
* put comment `"Conditionally define signal because AbortController throws error in development's strict mode"` above `L29`

(not sure if the second comment is true but that is what I assumed based on these comments)

minnakt (approved):
LGTM as long as failing tests pass!

PR #81, opened 2022-10-28
EVG-18085: Fix favicon path

minnakt:
EVG-18085

### Description 
Since `vite.config.ts` sets the `publicDir` to `<root>/public` by default, we don't need to include public in the path to the favicon. If you can't see the favicon in your browser after pulling this change, you probably need to hard refresh the tab.

### Screenshots
Also I made this green

Screen Shot 2022-10-28 at 11 53 42 AM

### Testing 
- I just looked at it 😳

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,7 @@
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <meta name="ui-version" content="%GIT_SHA%">
-  <link rel="shortcut icon" href="/public/favicon.ico" />
+  <link rel="shortcut icon" href="/favicon.ico" />
   <title>Parsley</title>
 </head>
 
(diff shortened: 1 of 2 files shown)


src/pages/LogDrop/FileDropper.tsx
```diff
@@ -61,7 +61,7 @@ const FileDropper: React.FC<FileDropperProps> = ({ onChangeLogType }) => {
 
   return (
     <Container>
-      <RedBorderBox>
+      <BorderBox>
```
khelif96, line 64: Nice

index.html
```diff
@@ -5,7 +5,7 @@
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <meta name="ui-version" content="%GIT_SHA%">
-  <link rel="shortcut icon" href="/public/favicon.ico" />
+  <link rel="shortcut icon" href="/favicon.ico" />
```
khelif96, line 8: Let's see a comprehensive set of unit and e2e tests to make sure this works. 😂

khelif96 (approved):

PR #82, opened 2022-10-28
EVG-18022: Upgrade LeafyGreen components to resolve emotion error

minnakt:
EVG-18022

### Description 
This PR upgrades both `emotion` and `LeafyGreen` components to get rid of the emotion error in the console.

### Screenshots
Bonus, this tooltip isn't wonky anymore after the upgrades:

Screen Shot 2022-10-28 at 3 40 24 PM

### Testing 
- Existing tests should pass


src/components/SearchBar/index.tsx
```diff
@@ -114,12 +114,13 @@ const SearchBar: React.FC<SearchBarProps> = ({
       <StyledInput
         ref={inputRef}
         aria-label="searchbar-input"
+        aria-labelledby="searchbar-input"
```
minnakt, line 117: I guess it's not great that we have both, but I need `aria-label` to avoid this console error and I need `aria-labelledby` to avoid a TypeScript error from LG

Screen Shot 2022-10-28 at 3 34 24 PM

khelif96 (approved):

PR #95, opened 2022-11-01
EVG-18195: Fix duplicate scrollbar issue

minnakt:
EVG-18195

### Description 
This PR removes the double scrollbar that was appearing in the `Y` direction. We were using `overflow` styling on the list container but we only really need `overflow-x` to enable horizontal scrolling.

Note that while there will no longer be two scrollbars, the scrollbars will still be visually different between mouse and trackpad. This is more a responsibility for the user, as they will need to change their scrolling settings to the following under General in System Preferences:

Screen Shot 2022-11-01 at 2 53 01 PM

Perhaps we can put this in a FAQ somewhere, although it is a Mac specific setting.

### Screenshots
- N/A

### Testing
- Fixed stories which had been showing incorrect behavior when applying wrap and expanding rows.
- Update e2e tests

### DIFF
--- a/src/components/LogPane/index.tsx
+++ b/src/components/LogPane/index.tsx
@@ -28,14 +28,24 @@ const LogPane: React.FC<LogPaneProps> = ({
   wrap,
   ...rest
 }) => {
-  const { listRef, matchingLines, prettyPrint } = useLogContext();
+  const { listRef, expandedLines, matchingLines, prettyPrint } =
+    useLogContext();
 
   const [expandableRows] = useQueryParam(QueryParams.Expandable, true);
+  const stringifiedExpandedLines = expandedLines.toString();
 
   useEffect(() => {
     cache.clearAll();
     listRef.current?.recomputeRowHeights();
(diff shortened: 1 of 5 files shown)


src/components/LogPane/index.tsx
```diff
@@ -28,22 +28,32 @@ const LogPane: React.FC<LogPaneProps> = ({
...
+    useLogContext();
 
   const [expandableRows] = useQueryParam(QueryParams.Expandable, true);
+  const stringifiedExpandedLines = expandedLines.toString();
 
   useEffect(() => {
     cache.clearAll();
     listRef.current?.recomputeRowHeights();
-  }, [listRef, cache, wrap, matchingLines, expandableRows, prettyPrint]);
+  }, [
```
khelif96, line 40: I'm noticing a strange jump when i expand a line now. 

https://user-images.githubusercontent.com/4605522/199344011-2a76d66f-8e34-4219-af68-047b2183300e.mov

khelif96 (changes requested):

sophstad (approved):

khelif96 (approved):
TY

PR #98, opened 2022-11-08
EVG-17668 Create prod deploy script for Parsley

khelif96:
EVG-17688

### Description 
Creates a script akin to the one in spruce that deploys parsley to prod by pushing a git tag and having evergreen do all the work. Also provides a fallback to deploy locally that does the same thing and sends an email.

The link to parsley production is  http://parsley.mongodb.com


scripts/deploy-production.js
```diff
@@ -0,0 +1,92 @@
...
+    console.log(commitMessages);
+  } else {
+    // If there are no commit messages, ask the user if they want to deploy anyway
+    const response = await prompts({
+      type: "confirm",
+      name: "value",
+      message: "No new commits. Do you want to deploy anyway?",
+      initial: false,
+    });
+    shouldDeployLocal = response.value;
```
minnakt, line 22: If the user says no at this point, should we exit the script? Otherwise they get prompted again about deploying:

Screen Shot 2022-11-09 at 11 20 24 AM

scripts/deploy-production.js
```diff
@@ -0,0 +1,92 @@
...
+        console.log(err);
+        console.log("Creating tag failed. Aborting.");
+      }
+    }
+  }
+};
+
+const createNewTag = () =>
+  new Promise((resolve, reject) => {
+    exec("yarn version --new-version patch", (err, stdout, stderr) => {
```
minnakt, line 58: I think you can remove this!
```suggestion
    exec("yarn version --new-version patch", (err, stdout) => {
```

package.json
```diff
@@ -29,10 +30,11 @@
     "start": "vite",
     "verify-backend": "./scripts/verify-backend.sh",
     "storybook": "start-storybook -p 6006",
-    "storybook:build": " . ./scripts/appVersion.sh && env-cmd -e local build-storybook",
+    "storybook:build": " . ./scripts/app-version.sh && env-cmd -e local build-storybook",
     "test": "jest --watchAll=false",
     "test:watch": "jest --watch --verbose",
-    "upload-source-maps": "node ./scripts/upload-source-maps.js"
+    "upload-source-maps": "node ./scripts/upload-source-maps.js",
+    "postversion": "git push upstream && git push upstream --tags && git push && git push --tags"
```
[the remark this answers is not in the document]
khelif96: postVersion automatically runs a command after we update the app version. So this handles pushing the tags

scripts/deploy-production.js
```diff
@@ -0,0 +1,92 @@
+const prompts = require("prompts");
+const { exec } = require("child_process");
+
```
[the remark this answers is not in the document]
khelif96: Nope you are right added!

scripts/deploy-production.js
```diff
@@ -0,0 +1,126 @@
...
+const localDeployScript =
+  "yarn build:production && env-cmd -e production yarn deploy:do-not-use && env-cmd -e production ./scripts/email.sh";
```
minnakt, line 5: seems like this should say `yarn build:prod`

Also wanted to mention that the `cmdrc` field in Spruce is called `prod`, but in Parsley it is called `production`. If you want to keep it as is that's fine with me though, as you would have to change it in multiple places

scripts/deploy-production.js
```diff
@@ -0,0 +1,126 @@
...
+      name: "value",
+      message: "No new commits. Do you want to deploy anyway?",
+      initial: false,
+    });
+    shouldDeployLocal = response.value;
+    if (!shouldDeployLocal) {
+      return;
+    }
+  }
+
```
minnakt, line 36: It might be slightly more readable if we format this part with just `if` statements rather than an `if-else` block. Something like this?

```javascript
  const commitMessages = await getCommitMessages();
  let shouldDeployLocal = false;

  if (commitMessages.length === 0) {
    const response = await prompts({
      type: "confirm",
      name: "value",
      message: "No new commits. Do you want to deploy anyway?",
      initial: false,
    });
    shouldDeployLocal = response.value;
    if (!shouldDeployLocal) {
      return;
    }
  }

  // Print all commits between the last tag and the current commit
  if (commitMessages.length > 0) {
    console.log("Commit messages:");
    console.log(commitMessages);
  }
```

.evergreen.yml
```diff
@@ -152,6 +171,34 @@ functions:
...
+        export DEPLOYS_EMAIL=${DEPLOYS_EMAIL}
+        export EXECUTION=${execution}
+        bash scripts/email.sh
+
+  setup-credentials:
+    command: shell.exec
+    params:
+      working_dir: parsley
+      script: |
+        echo "Generating .cmdrc.json"
```
minnakt, line 189: Should this say `.env-cmdrc.json`?

scripts/email.sh
```diff
@@ -0,0 +1,103 @@
...
+
+COMMAND="$EVERGREEN $CREDENTIALS notify email -f $AUTHOR_EMAIL -r $DEPLOYS_EMAIL -s "
+COMMAND+="'"
+COMMAND+="$DATE"
+COMMAND+=" $TITLE"
+COMMAND+="'"
+COMMAND+=" -b"
+COMMAND+=" '"
+COMMAND+="$BODY_HTML"
+COMMAND+="'"
```
minnakt, line 100: very nit but I feel like this could be condensed more

minnakt (approved):
my favorite part about this PR is how we didn't have to install an external module to check git stuff (like how we install `simple-git` in Spruce) 😎

PR #99, opened 2022-11-08
EVG-18235 Fix bug causing ScrollToLine not to run on initial render.

khelif96:
EVG-18235

### Description 
On initial page load with a selectedLine in the URL parsley would not scroll to the file.

### DIFF
--- a/src/components/LogWindow/index.tsx
+++ b/src/components/LogWindow/index.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect } from "react";
 import styled from "@emotion/styled";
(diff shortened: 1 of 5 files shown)


src/components/LogWindow/index.tsx
```diff
@@ -48,9 +48,11 @@ const LogWindow: React.FC<LogWindowProps> = ({ logType, isUploadedLog }) => {
     ? new RegExp(highlights.join("|"), "i")
     : undefined;
 
-  const [initialScrollIndex] = useState(
-    findLineIndex(processedLogLines, selectedLine)
```
khelif96, line 52: processedLogLines is not defined on the initial load.

PR #100, opened 2022-11-08
EVG-18235 Fix bug where scrollToLine would not work on load

khelif96:
EVG-18235

### Description 
ProcessedLogLines was not defined when we initially load the window so the scrollToLine would not work.


minnakt (approved):
Thanks, I like this approach a lot better than the other PR 😇

I have a feeling that we can remove the `hasLogs` guards in `LogWindow`, because it seems like `LogWindow` is only rendered if `hasLogs === true` (based on the parent components).

PR #104, opened 2022-11-17
EVG-18315 Hide expandable rows instead of removing filters

khelif96:
EVG-18315

### Description 
Expandable Rows should just hide the buttons instead of disabling filtering.

### DIFF
--- a/src/utils/filterLogs/filterLogs.test.ts
+++ b/src/utils/filterLogs/filterLogs.test.ts
@@ -25,7 +25,7 @@ describe("filterLogs", () => {
     ).toStrictEqual([0, 1, 2, 3, 4, 5, 6, 7]);
   });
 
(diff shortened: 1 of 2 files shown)


src/utils/filterLogs/filterLogs.test.ts
```diff
@@ -25,7 +25,7 @@ describe("filterLogs", () => {
     ).toStrictEqual([0, 1, 2, 3, 4, 5, 6, 7]);
   });
 
-  it("should return the log lines as is if expandable rows are turned off", () => {
+  it("should not return collapsed lines if expandedRows is turned off", () => {
```
minnakt, line 28: nit on wording
```suggestion
  it("should hide collapsed rows if expandableRows is turned off", () => {
```

minnakt (approved):
oh yeah! 🔥

PR #105, opened 2022-11-21
EVG-18347 Jump to first search result on search

khelif96:
EVG-18347

### Description 
Jump to the first matching line on search


minnakt (approved):
lgtm

PR #106, opened 2022-11-29
EVG-18376 Pressing enter key should paginate forward

khelif96:
EVG-18376

### Description 
Pressing enter now paginates forward. Also updated the shortcuts modal with new keys.

### Screenshots
image


src/components/ShortcutModal/index.tsx
```diff
@@ -7,15 +7,27 @@ import { size, zIndex } from "constants/tokens";
...
+  {
+    keys: [
+      ["CTRL", "S"],
+      ["CMD", "S"],
+    ],
     description: "Toggle between search, filter, and highlight",
   },
-  { keys: ["]"], description: "Toggle the filter panel" },
-  { keys: ["N"], description: "Paginate forward to the next search result" },
+  { keys: [["]"]], description: "Toggle the filter panel" },
```
minnakt, line 24: Oops my bad I should have noticed this; let's rename this like in the wiki! I think we went with `side panel`?

minnakt (approved):

PR #107, opened 2022-11-29
EVG-18369 Add Bookmark usage analytics and improve usability

khelif96:
EVG-18369

### Description 
Add analytics for bookmark usage.
Add tooltip for clearing bookmark
Update tooltip copy for jira copy button
### Screenshots
image
image

### DIFF
--- a/src/analytics/logWindow/useLogWindowAnalytics.ts
+++ b/src/analytics/logWindow/useLogWindowAnalytics.ts
@@ -11,11 +11,11 @@ type Action =
   | { name: "Toggled Filter"; visible: boolean }
   | { name: "Added Bookmark" }
   | { name: "Navigated With Bookmark" }
-  | { name: "Deleted Bookmark" }
+  | { name: "Removed Bookmark" }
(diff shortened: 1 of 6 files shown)


src/analytics/logWindow/useLogWindowAnalytics.ts
```diff
@@ -9,6 +9,10 @@ type Action =
   | { name: "Added Filter"; filterExpression: string }
   | { name: "Deleted Filter"; filterExpression: string }
   | { name: "Toggled Filter"; visible: boolean }
+  | { name: "Added Bookmark" }
+  | { name: "Navigated With Bookmark" }
+  | { name: "Deleted Bookmark" }
```
minnakt, line 14: `Removed Bookmark` imo is better phrasing, "delete" sort of implies to me that the bookmark entered the void never to be seen again. "remove" implies to me that the bookmark could appear again in the future

minnakt (approved):
as a temporary solution for user confusion surrounding bookmarks / JIRA button, lgtm

PR #108, opened 2022-11-29
EVG-18399 Fix wrong assumption about ctx field causing some resmoke lines not to be parsed.

khelif96:
EVG-18399

### Description 
Fixed an invalid assumption about resmoke lines which caused some resmoke lines not to be parsed correctly. 
### Screenshots
Before
image
After:
image

### Testing 
Unit testing and testing the broken log file.
Deploy beta and look at this log before and after
https://parsley-beta.corp.mongodb.com/resmoke/dbd0cc5bf43b847fdacda224a3233811/test/172c0eb2eeb3da45da7f096afcbcd301?bookmarks=0,3074,3078,3378,3379,3380,3385,3393,3394,3427,3428,3430,3431,3441,3445,3446,3447,3448,3450,3477,3492,3493,3494,3495,3496,3514,3515,3518,3523,3564,5134&expandable=false&filters=100I%20dont%20match%20anything


src/utils/resmoke/resmoke.test.ts
```diff
@@ -20,6 +31,15 @@ describe("processResmokeLine", () => {
       "[js_test:startup_recovery_commit_transaction_before_stable_timestamp] MongoDB shell version v6.2.0-alpha-250-g46817bf"
     );
   });
+  it("should process odd resmoke lines", () => {
```
minnakt, line 34: maybe make this slightly more specific, like `"should correctly process resmoke lines where ctx uses :: format"` (not sure if that's the correct condition, but something like that)

minnakt (approved):
wow, look at this power user, using bookmarks + filters + expandable row settings to make a point

src/utils/resmoke/resmoke.test.ts
khelif96: Good catch this is a dupe of tan above test

PR #124, opened 2022-12-01
EVG-18430 Fix bug causing log to scroll to top after clicking bookmark

khelif96:
EVG-18430

### Description 
For some reason after clicking on a bookmark scrolling would automatically take you to the top of the log.


minnakt (commented):
I'm hesitant to approve this change without knowing exactly why this happens. I'm okay with this solution, but I think it'd be better if we know why this is occurring before making this change.

I tried putting [03b29a6](https://github.com/evergreen-ci/parsley/commit/03b29a6c56d4d8d534be57d502224cd427a51860) on beta. At this point in the commit history, we didn't have this weird behavior.

Then I put [1bb0c3e](https://github.com/evergreen-ci/parsley/commit/1bb0c3e9c092ec59edcfc3cfab257a43dee9072a) on beta. This commit does have the weird behavior, so it seems to be coming from this commit. Do you know why these changes would cause these weird jumps?

minnakt:
Writing comment for future readers: this is another problem with the virtualization library 😕. Interactions between the CellMeasurer and Grid code cause the list to repeatedly force scroll to `scrollToIndex`, when in actuality that should not be happening.

src/components/LogPane/index.tsx
```diff
@@ -37,6 +38,15 @@ const LogPane: React.FC<LogPaneProps> = ({
     listRef.current?.recomputeRowHeights();
   }, [listRef, cache, wrap, matchingLines, expandableRows, prettyPrint]);
 
+  useEffect(() => {
```
minnakt, line 41: If we're going to run a `useEffect` on component mount, then I don't think `initialScrollIndex` needs to be a state in the parent component anymore.

minnakt (changes requested):

src/components/LogPane/index.tsx
khelif96: I replaced it with a useMemo

minnakt (approved):
doing awesome stuff 🙏

PR #127, opened 2022-12-02
EVG-18441 Handle timestamps with negative offsets

khelif96:
EVG-18441

### Description 
Some logs featured negative timestamps which would not be detected by our regex this updates the offset regex to support it. 
### Screenshots
Before:
image
After:
image


minnakt (approved):
our local regex professional 😁

PR #131, opened 2022-12-06
EVG-18375: Render rows that contain empty strings

minnakt:
EVG-18375

### Description 
This PR makes it so that we render rows that contain empty strings. The reason they were not being rendered before is because we were only checking if `lineContent` is falsy.

#### Ideas
Maybe we can introduce an option when parsing the log to omit lines that only contain empty strings? We would need to know if users would actually find this helpful though.

### Screenshots
Screen Shot 2022-12-06 at 11 11 59 AM

### Testing 
- Added tests in `ResmokeRow.test.tsx`
- Added tests in `AnsiiRow.test.tsx`


khelif96 (approved):
Great find!

PR #137, opened 2022-12-13
EVG-18492: Fix drag and drop finickiness

minnakt:
EVG-18492

### Description 
This PR modifies the code in `FileDropper` to better adhere to the [documentation](https://react-dropzone.js.org/#section-opening-file-dialog-programmatically). I removed the `"Release to upload"` prompt because I couldn't get it to work with it, and I don't think it's really necessarily.

Also changed text input type to `"text"` to prevent the `"LeafyGreen recommends using SearchInput..."` warnings.

### DIFF
--- a/src/components/SearchBar/__snapshots__/SearchBar.stories.storyshot
+++ b/src/components/SearchBar/__snapshots__/SearchBar.stories.storyshot
@@ -79,7 +79,7 @@ exports[`storyshots Storyshots Components/SearchBar Default 1`] = `
           placeholder="optional, regexp to search"
           required={true}
           spellCheck={false}
-          type="search"
+          type="text"
           value=""
         />
         <div
--- a/src/pages/LogDrop/FileDropper.tsx
+++ b/src/pages/LogDrop/FileDropper.tsx
@@ -59,9 +59,11 @@ const FileDropper: React.FC<FileDropperProps> = ({ onChangeLogType }) => {
     }
   }, [ingestLines, logType, sendEvent, onChangeLogType]);
 
-  const { getRootProps, getInputProps, isDragActive, open } = useDropzone({
+  const { getRootProps, getInputProps, open } = useDropzone({
     onDrop,
     maxFiles: 1,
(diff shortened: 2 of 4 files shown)


src/pages/LogDrop/FileDropper.tsx
```diff
@@ -59,59 +59,59 @@ const FileDropper: React.FC<FileDropperProps> = ({ onChangeLogType }) => {
...
     onDrop,
     maxFiles: 1,
+    noClick: true,
+    noKeyboard: true,
   });
 
   return (
     <Container>
       <BorderBox>
         {hasDroppedLog ? (
```
sophstad, line 72: Just a UX note, and no worries if the scope of this is a little difficult and you'd rather not, but: it would be nice if the behavior were a little more defined if a user tries to drag and drop an additional log. I would probably expect either a clear indicator that the operation failed, or for the second log to replace the first. Right now nothing in particular happens.

minnakt: I updated the behavior! Lmk what you think!

src/components/SearchBar/__snapshots__/SearchBar.stories.storyshot
```diff
@@ -79,7 +79,7 @@ exports[`storyshots Storyshots Components/SearchBar Default 1`] = `
           placeholder="optional, regexp to search"
           required={true}
           spellCheck={false}
-          type="search"
+          type="text"
```
sophstad, line 82: It's annoying that we have to make this change 🙄 but I guess the warning is more annoying 😆

sophstad (approved):
Now That's What I Call a Great UX!

src/components/SearchBar/__snapshots__/SearchBar.stories.storyshot
minnakt: yeah hmmm the placement of the search icon on the `SearchInput` makes it not so great for a one-to-one replacement 😔

PR #138, opened 2022-12-14
EVG-16639: Record users' preferences in cookies

minnakt:
EVG-16639

### Description 
This PR makes it so that users' filter options are stored in cookies. However, the URL param always takes precedence over the cookie value.

I consolidated these filter options in `LogContext/index.tsx` because otherwise I would need to instantiate them with the cookie values every time.

### Screenshots
- N/A

### Testing 
- Added unit tests for each of the toggles


src/context/LogContext/index.tsx
```diff
@@ -245,24 +271,35 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
         dispatch({ type: "SET_SEARCH_TERM", searchTerm });
       },
+      setWrap,
       togglePrettyPrint: (v: boolean) => {
         setPrettyPrint(v);
         Cookie.set(PRETTY_PRINT_BOOKMARKS, v.toString(), { expires: 365 });
       },
     }),
     [
+      wrap,
```
minnakt, line 297: Actually I have a question. Does `wrap` truly need to be in the URL? If it always rendered solely based on the cookie, it would lead to a more consistent experience for users.

khelif96: I agree as well wrap seems more like a preference than a shareable value. @mvankeulen94  can you confirm?

minnakt: I talked to Maria about it and she said that there wasn't much of a use case for `wrap` being shareable, so we can turn off sharing for it. I'll update this PR with those changes

src/context/LogContext/index.tsx
```diff
@@ -245,24 +268,35 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
+      setWrap,
+      filterLogic,
+      setFilterLogic,
+      expandableRows,
+      setExpandableRows,
```
khelif96, line 299: How do you feel about grouping these into a `preferences` object?

khelif96:
nit. for the pr title `[EVG-16639: Record users' preferences in cookies](https://github.com/evergreen-ci/parsley/pull/138#top)

src/context/LogContext/index.tsx
minnakt: good idea :]

khelif96 (approved):
Great refactor!

PR #140, opened 2022-12-15
EVG-18245 Upgrade cypress to v12

khelif96:
EVG-18245


cypress.config.ts
```diff
@@ -7,8 +7,7 @@ export default defineConfig({
     supportFile: "cypress/support/index.ts",
     specPattern: "cypress/integration/**/*.ts",
     experimentalStudio: true,
-    experimentalSessionAndOrigin: true,
-    testIsolation: "legacy",
+    testIsolation: false,
```
minnakt, line 10: Is it worth it to file a ticket to do work to be able to turn this on? I do think it would be good to follow their best practices, even if the behavior is not what we are used to.

minnakt (approved):
lgtm as long as the failing tests are fixed!

cypress.config.ts
khelif96: While I do agree tests should be isolated across test suites. I don't necessarily agree with test isolation on a per-test basis for e2e tests. The nature of our apps involve setting up complex state through a series of interactions and it gives us the opportunity to make assertions along the way. Having to start on a clean slate for each test would require us to boilerplate and repeat all the necessary commands to get to the point of an assertion every single time. 

I added this to the UI roundtable agenda to discuss with the team.

PR #146, opened 2022-12-21
EVG-18624 Turn on auto_update for modules

khelif96:
EVG-18624


minnakt (approved):
good! great! :bowtie:

PR #148, opened 2022-12-27
EVG-18633 Avoid processing log files twice

khelif96:
EVG-18633

### Description 
Introduces a `useEffectDebugger` hook which performs the same functionality as a useEffect but will also log the dependencies that have changed.

While investigating the above hook I realized we were processing log files twice due to an unstable dependency that was being re-initialized whenever the context would update.


sophstad (approved):
Cool!

PR #150, opened 2022-12-28
EVG-18247 EVG-18244 Bump eslint and jest

khelif96:
EVG-18247
EVG-18244

### Description 
Bump dependencies for jest and eslint


.eslintrc.js
```diff
@@ -180,6 +181,7 @@ module.exports = {
         "jest/no-hooks": OFF,
         "jest/no-mocks-import": OFF,
         "jest/prefer-expect-assertions": OFF,
+        "jest/unbound-method": OFF,
```
minnakt, line 184: Is this error expected when the rule is turned on?

Screen Shot 2023-01-03 at 12 01 41 PM

khelif96: Fixed it.

.eslintrc.js
```diff
@@ -81,6 +81,7 @@ module.exports = {
       ERROR,
       { functions: false, variables: false },
     ],
+    "@typescript-eslint/unbound-method": OFF,
```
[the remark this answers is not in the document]
khelif96: Turned it back on and fixed the errors.  
This rule is likely included as one of the preset rules we are extending. This was likely added at some point after we upgraded the deps.

.eslintrc.js
```diff
@@ -180,6 +184,8 @@ module.exports = {
         "jest/no-hooks": OFF,
         "jest/no-mocks-import": OFF,
         "jest/prefer-expect-assertions": OFF,
+        "@typescript-eslint/unbound-method": OFF,
+        "jest/unbound-method": OFF,
```
minnakt, line 188: It is frustrating that the only reason we have to turn this on is because of `Cookie.get` 😔

minnakt (approved):

PR #161, opened 2023-01-09
EVG-18167: Refactor code around LogPane and Rows

minnakt:
EVG-18167

### Description 
This PR refactors code related to `LogPane` and `Row`s (`AnsiiRow`, `ResmokeRow`, `CollapsedRow`). This refactor was kind of tricky, and I'm open to any suggestions to improve it. The main things I wanted to do were:
* Make the props inherited by `CollapsedRow` distinct from other rows. It doesn't make sense that whenever we modify the props for `CollapsedRow`, we have to modify the props to the other rows as well, even though they will never use them.
* Reduce the amount of files we have to modify whenever we add a prop on `AnsiiRow` or `ResmokeRow` by not passing the data through `LogPane`.
* Remove tests that unnecessarily double test functionality, e.g. there's no reason to test bookmarking in `AnsiiRow` as that is not a feature of `AnsiiRow`, that is a feature of `BaseRow`.

### Screenshots
* N/A

### Testing 
* Removed tests that were double testing functionality in `AnsiiRow` and `ResmokeRow`
* Modified other tests files to accommodate the changes


src/components/BookmarksBar/index.tsx
```diff
@@ -24,25 +23,21 @@ const BookmarksBar: React.FC<BookmarksBarProps> = ({
...
+    setTimeout(() => {
+      if (bookmarks.length === 0 && lineCount !== 0) {
+        if (lineCount === 1) {
+          setBookmarks([0]);
+        } else {
+          setBookmarks([0, lineCount - 1]);
+        }
       }
-    }
+    }, 0);
```
minnakt, line 40: One of the things I did in this refactor was centralize the URL params in `LogContext`, which causes this issue. :/ 

I wonder if I shouldn't put them in the `LogContext`, but having them in one place can also be useful... Another option is to move this URL update into `LogContext`, but then it's not testable as something that happens in `BookmarksBar`.

khelif96: I'm not to big of a fan of this approach if the only reason is to appease jest.

khelif96 (commented):
Just a quick skim before i leave. I didn't get the chance to fully review it or test it out. But initially had two comments.

minnakt:
Update about what this PR does now:
- `CollapsedRow` props are distinct from other `Row` props.
- `LogPane` no longer passes data through the `RowRenderer`.
- Remove redundant tests for `AnsiiRow` and `ResmokeRow`.

I would also be okay with postponing this work if we feel like there's not enough information / understanding yet, because it seems like this could affect future projects. In that case I would close this PR and release the ticket to the backlog.

src/components/LogRow/BaseRow/index.tsx
```diff
@@ -40,34 +40,35 @@ const BaseRow = forwardRef<any, BaseRowProps>((props, ref) => {
...
-  const highlighted = highlightedLine === index;
+  const shared = shareLine === lineNumber;
+  const bookmarked = bookmarks.includes(lineNumber);
+  const highlighted = searchLine === index;
 
-  // Clicking a line should select or deselect the line.
+  // Clicking link icon should set or unset the share line.
   const handleClick = () => {
     if (shared) {
       setShareLine(undefined);
```
khelif96, line 74: I can't select the line below this but wrapping handleDoubleClick with a `useCallback` prevents an entire render.

src/components/LogWindow/index.tsx
```diff
@@ -16,27 +15,11 @@ const LogWindow: React.FC<LogWindowProps> = ({ logType, isUploadedLog }) => {
...
-    range,
-    searchState,
   } = useLogContext();
-  const { prettyPrint, wrap } = preferences;
-  const { searchTerm } = searchState;
-
-  const [highlights] = useHighlightParam();
-  const highlightRegex = highlights.length
-    ? new RegExp(highlights.join("|"), "i")
-    : undefined;
```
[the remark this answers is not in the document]
minnakt: I’m questioning if this is true, since `RowRenderer` is actually a singular component with a for-loop inside of it. Since the highlight regex is declared outside of the for-loop, I think it only gets defined once. I checked this manually to the best of my ability, but let me know if I missed something.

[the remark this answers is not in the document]
khelif96: Great call I hadn't noticed that!

khelif96 (approved):
Great job on this! 🚀

PR #162, opened 2023-01-09
EVG-18397: Always enable Expand 5 button

minnakt:
EVG-18397

### Description 
This PR makes it so the `Expand 5 Above and Below` button is always enabled. It also modifies `mergeIntervals` function to be slightly more robust.

### Screenshots
Screen Shot 2023-01-09 at 12 11 36 PM

### Testing 
- Added test in `utils/expandedLines/expandedLines.test.ts`

### DIFF
--- a/src/components/LogRow/CollapsedRow/index.tsx
+++ b/src/components/LogRow/CollapsedRow/index.tsx
@@ -32,13 +32,17 @@ const CollapsedRow = forwardRef<any, CollapsedRowProps>((props, ref) => {
     numCollapsed !== 1 ? `${numCollapsed} lines skipped` : "1 line skipped";
 
   const expandFive = () => {
-    startTransition(() =>
-      expandLines([
-        [start, start + (SKIP_NUMBER - 1)],
-        [end - (SKIP_NUMBER - 1), end],
-      ])
-    );
-    sendEvent({ name: "Expanded Lines", lineCount: 5, option: "Five" });
+    if (2 * SKIP_NUMBER < numCollapsed) {
+      startTransition(() =>
+        expandLines([
+          [start, start + (SKIP_NUMBER - 1)],
+          [end - (SKIP_NUMBER - 1), end],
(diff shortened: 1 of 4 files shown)


src/components/LogRow/CollapsedRow/index.tsx
```diff
@@ -27,25 +30,23 @@ const CollapsedRow = forwardRef<any, CollapsedRowProps>((props, ref) => {
...
+    if (2 * SKIP_NUMBER < numCollapsed) {
+      startTransition(() =>
+        expandLines([
+          [start, start + (SKIP_NUMBER - 1)],
+          [end - (SKIP_NUMBER - 1), end],
+        ])
+      );
+      sendEvent({ name: "Expanded Lines", lineCount: 5, option: "Five" });
+    } else {
+      expandAll();
```
khelif96, line 44: I think it still makes sense to track this as Expanded Five since that's the button they clicked.

src/utils/expandedLines/expandedLines.test.ts
```diff
@@ -11,6 +11,16 @@ describe("mergeIntervals", () => {
...
   });
 
+  it("merges overlapping intervals when end + 1 > start", () => {
+    expect(
+      mergeIntervals([
+        [0, 5],
+        [3, 15],
+        [7, 28],
+      ])
+    ).toStrictEqual([[0, 28]]);
```
khelif96, line 21: What scenario would this occur in? I'm glad we are handling this but I'm wondering if there is a case where this could happen.

khelif96 (approved):
Works great had a few questions/ comments

src/utils/expandedLines/expandedLines.test.ts
minnakt: in theory, it should never happen. But I figured it would be good to cover all bases, especially since it has no impact on current functionality :]

PR #163, opened 2023-01-10
EVG-18493 Fix bug causing strange log rendering for ansi logs

khelif96:
EVG-18493

### Description 
I think declaring AnsiUp as a singleton was causing it to get confused and would cause it to highlight the wrong lines sometimes. 

### Screenshots

https://user-images.githubusercontent.com/4605522/211610856-cbf0091e-2481-422c-96f5-10814f395aae.mov

### Testing 
Scrolling no longer causes log line color corruption.  I can't think of a clear way to introduce any tests for this since it it was flaky.


minnakt (approved):
lgtm, can confirm it works for me 🪱

PR #167, opened 2023-01-18
EVG-18680: Replace selectedLine with shareLine param

minnakt:
EVG-18680

### Description 
This PR makes it so that the `selectedLine` param is replaced with the `shareLine` param. `shareLine` is a better name for this query param because it makes it clear what it does.

I opened [EVG-18748](https://jira.mongodb.org/browse/EVG-18748) to delete the code around `selectedLine` at a later date.

### Screenshots
- No visible changes.

### Testing 
- Added test in `routes.ts` (Cypress)

### Note 
This ticket also includes updating the links in Evergreen [here](https://github.com/evergreen-ci/evergreen/blob/main/model/task/task.go#L483-L486). However, this PR will need to be merged and deployed **_before_** that change is made.

### DIFF
--- a/src/components/BookmarksBar/index.tsx
+++ b/src/components/BookmarksBar/index.tsx
@@ -7,7 +7,7 @@ import { useLogWindowAnalytics } from "analytics";
 import Icon from "components/Icon";
 import { QueryParams } from "constants/queryParams";
 import { size } from "constants/tokens";
-import { useQueryParam, useQueryParams } from "hooks/useQueryParam";
+import { useQueryParam } from "hooks/useQueryParam";
 import { findLineIndex } from "utils/findLineIndex";
 
 const { gray, green } = palette;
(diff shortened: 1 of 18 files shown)


src/components/BookmarksBar/index.tsx
```diff
@@ -24,16 +24,33 @@ const BookmarksBar: React.FC<BookmarksBarProps> = ({
...
+  // This code can be deleted in EVG-18748.
+  useEffect(() => {
+    if (selectedLine) {
+      setSearchParams({
+        ...searchParams,
+        shareLine: selectedLine,
+        selectedLine: undefined,
+      });
+    }
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
khelif96, line 52: Can we move this to a more generic location? Somewhere like `LogView` or `LogWindow`? Several parts of the app use `selectedLine` so it seems odd to me to handle this transformation in a random component.

khelif96 (changes requested):
Looks like the automated scroll doesn't work if selectedLine is passed in.

https://user-images.githubusercontent.com/4605522/213782728-7e4ddb29-bc7b-499f-8455-fe49f69388c1.mov

src/components/BookmarksBar/index.tsx
minnakt: I moved it to `LogContext`, which may be a little awkward but works the best because the URL param gets replaced while the log is still downloading. This allows the jump to line functionality to work again, because `shareLine` is finished being overwritten by the time the log loads. Since this is only a temporary solution that will eventually be deleted, I think it's okay to have it in `LogContext`.

khelif96 (approved):

PR #169, opened 2023-01-20
EVG-18491: Fix laggy loading animation on Chrome

minnakt:
EVG-18491

### Description 
This PR fixes the laggy animation when loading a large log file on Chrome. Initially I (incorrectly) thought we might have to move the log processing off of the main thread, but then I noticed how the animation wouldn't lag at all on Safari but would perform pretty terribly on Chrome.

I think what's happening is that Chrome doesn't handle animating SVGs well (the Parsley icon gets rendered as an SVG.) Wrapping the SVG in a `div` and applying the animation on the `div` got rid of the lag.

### Screenshots
(On Chrome)

https://user-images.githubusercontent.com/47064971/213779971-05a4d467-b197-4667-8694-b67135840129.mov

### Testing 
- Manually looking


khelif96 (approved):
Wow AMAZING find!

PR #172, opened 2023-01-24
EVG-18419: Improve search workflow

minnakt:
EVG-18419

### Description 
This PR implements the following changes:
* Removes the "Search" option. By default, the text input will always search.
* Makes it so that pressing `Cmd + Shift + Enter` will submit the filter or highlight.
* Adds a guide cue to explain the new functionality.

I downloaded `lodash.debounce` because it has a native `cancel()` function and I didn't really want to spend the time figuring it out for our custom function 🧍 We have to cancel the debounce request or else what can happen is that a user will submit a filter/highlight and then an unexpected search will appear a few seconds later.

### Screenshots
(Maybe the message should be more generic so it could be understood by new users...)

Screen Shot 2023-01-24 at 3 06 09 PM

### Testing 
- Added tests in `SearchBarGuideCue.test.tsx`
- Modified Cypress tests to accommodate for changes


cypress/integration/resmokeLogs/resmoke_filtering.ts
```diff
@@ -199,7 +197,7 @@ describe("Filtering", () => {
     const filter = "doesNotMatchAnything";
 
     beforeEach(() => {
-      cy.login();
+      cy.setCookie("has-opened-drawer", "false");
```
khelif96, line 200: Can we turn this into a reusable cypress command?

src/components/Search/SearchBar/index.tsx
```diff
@@ -66,34 +62,26 @@ const SearchBar: React.FC<SearchBarProps> = ({
...
   const handleOnSubmit = () => {
-    if (inputRef.current) {
-      inputRef.current.blur();
-    }
-    if (isFilter || isHighlight) {
-      setInput("");
-    }
+    // Cancel the debounce request and clear input.
+    debounceSearch.cancel();
+    inputRef.current?.blur();
```
khelif96, line 74: There's an interesting side effect of this change. Paginating by pressing enter is no longer as intuitive. Since we no longer blur the input pressing enter no longer paginates the search since the input is still focused. I think we would want to fix that since it is frustrating.

cypress/support/index.ts
```diff
@@ -87,3 +87,9 @@ declare global {
     }
   }
 }
+
+beforeEach(() => {
```
khelif96, line 91: nice 👍

src/components/Search/SearchBarGuideCue/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [openGuideCue, setOpenGuideCue] = useState(
+    Cookie.get(HAS_SEEN_SEARCHBAR_GUIDE_CUE) !== "true"
+  );
+
+  const onHideCue = () => {
+    Cookie.set(HAS_SEEN_SEARCHBAR_GUIDE_CUE, "true", { expires: 365 });
+    setOpenGuideCue(false);
+  };
+
+  return (
```
khelif96, line 29: nit. but not necessary. Storybook story incorporating this would be nice.

src/components/Search/SearchBar/index.tsx
```diff
@@ -66,34 +62,26 @@ const SearchBar: React.FC<SearchBarProps> = ({
     { disabled, ignoreFocus: true }
   );
 
+  const debounceSearch = useRef(
+    debounce((value: string) => {
+      onChange(value);
+    }, 1000)
+  ).current;
```
khelif96, line 69: I'm confused about what this useRef is for?

src/components/Search/index.tsx
```diff
@@ -49,19 +48,22 @@ const Search: React.FC = () => {
...
   };
 
+  const [containerRef, setContainerRef] = useState<HTMLDivElement | null>(null);
+
   return (
-    <Container>
+    <Container ref={(el) => setContainerRef(el)}>
+      {hasLogs && containerRef && (
+        <SearchBarGuideCue containerRef={containerRef} />
+      )}
```
khelif96, line 62: You can use `useRef` here. Although I would argue the hook should be at the top of the component.  :) 
```ts
  const containerRef = useRef<HTMLDivElement>(null);
  return (
    <Container ref={containerRef}>
      {hasLogs && containerRef.current && (
        <SearchBarGuideCue containerRef={containerRef.current} />
      )}
...
```

src/components/Search/SearchBar/index.tsx
khelif96: If its to maintain a stable reference to the debounce function you can use [`useCallback`](https://beta.reactjs.org/reference/react/useCallback)

src/components/Search/index.tsx
minnakt: ty! 🙏

src/components/Search/SearchBar/index.tsx
minnakt: Yeah, I noticed that but figured it would be remedied by [EVG-18576](https://jira.mongodb.org/browse/EVG-18576). I ended up just doing the ticket in this PR.

src/components/Search/SearchBar/index.tsx
minnakt: I needed a reference to `debounceSearch` to call the `cancel()` function on it, and I couldn't figure out how to do this with `useCallback` ☹️

src/components/Search/SearchBar/index.tsx
minnakt: Oh also, I realized that `Shift + Enter` wouldn't work for submitting since that is what's used to paginate backwards. I changed it to `Cmd + Shift + Enter`.

cypress/support/index.ts
```diff
@@ -62,15 +62,20 @@ declare global {
...
        */
       toggleDrawer(): void;
       /**
-       * Custom command to open and close the Details Panel.
-       * @example cy.toggleDetailsPanel(true)
+       * Custom command to unset the drawer cookie.
+       * @example cy.unsetDrawerCookie()
        */
-      toggleDetailsPanel(open: boolean): void;
+      unsetDrawerCookie(): void;
```
khelif96, line 78: What do you think about `resetDrawerState`?

src/components/Search/SearchBar/index.tsx
```diff
@@ -90,6 +93,17 @@ const SearchBar: React.FC<SearchBarProps> = ({
...
   };
 
+  const onKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+    const commandKey = e.ctrlKey || e.metaKey;
+    if (e.key === "Enter" && commandKey && isValid) {
+      handleOnSubmit();
+    } else if (e.key === "Enter" && e.shiftKey) {
+      paginate(DIRECTION.PREVIOUS);
+    } else if (e.key === "Enter") {
+      paginate(DIRECTION.NEXT);
```
khelif96, line 103: you can use the `ignoreFocus` option in the original `useKeyboardShortcut` to allow pagination if the input is focused.

khelif96 (changes requested):
Works great so far just two more comments

src/components/Search/SearchBar/index.tsx
khelif96: No need to reimplement pagination here.

minnakt: I think it's important to tie pagination to the search bar explicitly, because we don't want to allow pagination from any text input:

https://user-images.githubusercontent.com/47064971/215210402-5b2caf0c-5dce-40b0-913d-56eb5210539c.mov

I do think it would be possible to delete the keyboard shortcuts for `Enter` and `Shift + Enter` though... I don't think they really needed under this new functionality.

khelif96: hmm good point. I don't like that the logic for pagination is tied to this input however since they are applying on separate functions. Would a wrapper component that encompasses both make sense? Or maybe refactor `useKeyboardShortcut` so that it accepts an `allowList` of accepted inputs.

khelif96 (approved):
Great Job!

PR #174, opened 2023-01-27
EVG-18664: Use Electron browser with Cypress

sophstad:
EVG-18664

### Description 

- OOM errors appear to have gone away for now, but [v12.4.0 added](https://docs.cypress.io/guides/references/changelog#12-4-0) an `experimentalMemoryManagement` flag that we can introduce if they pop up again
- Electron browser no longer requires enabling clipboard permissions
- Update `copyToClipboard` function since [`writeText()` is  async](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText)

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -163,16 +163,6 @@ functions:
     params:
       directory: parsley
 
-  install-chrome:
-    command: shell.exec
-    params:
-      working_dir: parsley
-      script: |
-        cd /tmp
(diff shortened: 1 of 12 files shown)


minnakt (approved):
for the OOM error mentioned in the description, does that mean we can remove `numTestsKeptInMemory` in `cypress.config.ts`? But I also wonder if it could be because I switched the resmoke log we were filtering on 😅

khelif96 (approved):
12.4.1 👀

PR #187, opened 2023-02-03
EVG-2709 Support highlight logs with multiple colors

khelif96:
EVG-2709

### Description 
* Supports highlighting in multiple colors. I took the colors we use for resmoke syntax highlighting and swapped them to use their light variants.
*  I also updated the hiding of the console.log messages to only work on CI. 
* Created storybook stories for sidenav components

### Screenshots
image

### Testing 
Unit, e2e tests, storybook story


src/components/SidePanel/NavGroup/HighlightNavGroup/index.tsx
```diff
@@ -28,7 +29,7 @@ const HighlightNavGroup: React.FC = () => {
       items={highlights}
       navGroupTitle="Highlighted Terms"
     >
-      {highlights.map((highlight) => (
+      {highlights.map((highlight, index) => (
```
minnakt, line 32: I'm somewhat wondering if we should do something more robust than an array. I was thinking about in the future, if we would want to allow users to change the color and provide some colors to choose from.

Screen Shot 2023-02-07 at 3 46 49 PM

minnakt (commented):
I'm still thinking about the regex logic, but I left some comments to start! I'm also thinking there's maybe a larger conversation to be had about how to resolve highlight clashes.

src/components/Highlight/index.tsx
```diff
@@ -1,15 +1,30 @@
...
-const Highlight = styled.mark<{ "data-cy"?: string }>`
-  background-color: ${red.light2};
+const Highlight = styled.mark<{ "data-cy"?: string; color?: string }>`
+  background-color: ${({ color }) => color};
   font-weight: bold;
 `;
 
 Highlight.defaultProps = {
   "data-cy": "highlight",
+  color: red.light2,
```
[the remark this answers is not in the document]
khelif96: Yes Since we have a finite amount of colors we can use anything else should just be the default color.

src/components/SidePanel/NavGroup/HighlightNavGroup/index.tsx
khelif96: As awesome as that looks I would push back against it at least for this first iteration since that would dramatically increase implementation complexity for little benefit IMO. Right now the colors are independent of app state doing it that way would require us to keep track of what colors are associated with what highlights.

src/components/SidePanel/NavGroup/FilterNavGroup/FilterNavGroup.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+        visible: true,
+      },
+    ]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+  const clearExpandedLines = () => actions("clearExpandedLines");
+  return <FilterNavGroup clearExpandedLines={clearExpandedLines} {...args} />;
+};
+
+export const Default: StoryObj<typeof FilterNavGroup> = {
```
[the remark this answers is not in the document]
khelif96: I assume its just the styling of the NavGroup Option when its not inside a sidenav

src/components/SidePanel/NavGroup/HighlightNavGroup/index.tsx
```diff
@@ -37,7 +38,14 @@ const HighlightNavGroup: React.FC = () => {
           >
             <Icon glyph="X" />
           </IconButton>
-          <Overline>{highlight}</Overline>
+          <Overline>
```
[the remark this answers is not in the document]
khelif96: Great catch fixed

src/components/Highlight/index.tsx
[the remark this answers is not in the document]
minnakt: What do you think about looping through the available colors using the modulo operator?

src/components/SidePanel/NavGroup/HighlightNavGroup/index.tsx
minnakt: Okay, that makes sense to me to not to support it for that use case at the current time 👍

I wanted to point out one other scenario where this might come into question, which is when we delete highlights. Deleting highlights will change the colors in view, which I'm not sure how users will feel about. (This could be more of a product/design question, but wanted to flag.)

https://user-images.githubusercontent.com/47064971/218574955-e68a2b49-ca2d-46e0-8d81-ec6f91e71c3c.mov

src/components/Highlight/index.tsx
```diff
@@ -1,15 +1,30 @@
...
 };
 
+const highlightColorList = [
+  green.light1,
+  blue.light2,
+  red.light3,
```
minnakt, line 19: This color might be a candidate for removal, as it can't be seen when a line is the active search line

src/components/Highlight/index.tsx
```diff
@@ -1,15 +1,30 @@
...
+  yellow.light2,
+  green.light2,
+  purple.light2,
+  blue.base,
+  green.light3,
+  red.light2,
+  yellow.light3,
```
minnakt, line 26: This color might be a candidate for removal, as it can't be seen when a line is bookmarked

src/components/LogRow/BaseRow/index.tsx
```diff
@@ -142,8 +143,9 @@ const ProcessedBaseRow: React.FC<ProcessedBaseRowProps> = memo((props) => {
     let render = children;
     if (searchTerm) {
       // escape the matching string to prevent XSS
-      render = render.replace(
+      render = highlighter(
```
minnakt, line 146: Might be something funky going on with the replace regex, `Script` should be red and `deploy` should be purple but they are both a bright yellow not defined in the LeafyGreen system

Screen Shot 2023-02-13 at 5 48 48 PM

minnakt (commented):
if any of my comments are indicative of a larger issue and need to spun off to another ticket(s), that would be completely fine 👍

src/components/LogRow/BaseRow/index.tsx
khelif96: Looks like the color was being passed in as undefined the yellow is the default color for the `<mark> tag </mark>`

src/components/SidePanel/NavGroup/FilterNavGroup/FilterNavGroup.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+  render: (args) => <Story {...args} />,
+};
```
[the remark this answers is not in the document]
khelif96: Done

src/components/SidePanel/NavGroup/HighlightNavGroup/index.tsx
khelif96: I think this is an ok compromise especially given the low usage of this feature.

src/components/Highlight/Highlight.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  ),
+};
+const meta: Meta<typeof Highlight> = {
+  component: Highlight,
+  argTypes: {
+    color: {
+      control: "select",
+      options: highlightColorList,
+    },
+  },
```
minnakt, line 33: I tried interacting with this, but I'm not sure it does anything

```suggestion

```

src/components/Highlight/index.tsx
```diff
@@ -1,15 +1,29 @@
...
-const Highlight = styled.mark<{ "data-cy"?: string }>`
-  background-color: ${red.light2};
+const Highlight = styled.mark<{ "data-cy"?: string; color?: string }>`
+  background-color: ${({ color }) => color};
   font-weight: bold;
 `;
 
 Highlight.defaultProps = {
   "data-cy": "highlight",
+  color: red.light2,
```
minnakt, line 13: should this be `color` or `background-color`, based on the styling above it?

src/components/LogRow/types.ts
```diff
@@ -10,7 +10,6 @@ interface LogRowProps extends BaseRowProps {
   scrollToLine: (lineNumber: number) => void;
 
   highlightRegex?: RegExp;
-  highlights?: RegExp;
```
minnakt, line 13: ty

minnakt (approved):
great job!

src/components/Highlight/Highlight.stories.tsx
khelif96: It does 
image

PR #188, opened 2023-02-03
EVG-18872: Leave breadcrumb on error page

minnakt:
EVG-18872

### Description 
This PR leaves a breadcrumb on the error page. The breadcrumb will inform us if a user has seen the error page or not.

I put the breadcrumb in `componentDidCatch` based on this page about error boundaries ([link](https://reactjs.org/docs/error-boundaries.html)):

Screen Shot 2023-02-02 at 9 52 50 AM

Note that I don't think you can see `console.debug` in Chrome by default (have to turn on verbose logging level).

### Testing 
- Tested manually by throwing error when accessing a value that didn't exist (e.g. `filters[2].caseSensitive`) and checked that the message was console logged

### DIFF
--- a/src/components/ErrorBoundary/index.tsx
+++ b/src/components/ErrorBoundary/index.tsx
@@ -37,7 +37,7 @@ class DefaultErrorBoundary extends Component<
   }
 
   componentDidCatch(error: Error, errorInfo: any) {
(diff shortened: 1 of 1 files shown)


khelif96 (approved):
LGTM

PR #199, opened 2023-02-14
EVG-18958 Lower overscan count to 50

khelif96:
EVG-18958

### Description 
This is an exploratory pr to see if lowering the Overscan count causes people to report less log jumpiness.  See linked ticket for more info

### DIFF
--- a/src/components/LogPane/index.tsx
+++ b/src/components/LogPane/index.tsx
@@ -70,7 +70,7 @@ const LogPane: React.FC<LogPaneProps> = ({
           }}
           deferredMeasurementCache={cache}
           height={height}
-          overscanRowCount={200}
+          overscanRowCount={50}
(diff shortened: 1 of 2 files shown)


src/components/LogPane/index.tsx
```diff
@@ -70,7 +70,7 @@ const LogPane: React.FC<LogPaneProps> = ({
           }}
           deferredMeasurementCache={cache}
           height={height}
-          overscanRowCount={200}
+          overscanRowCount={50}
```
minnakt, line 73: One thing that I might also recommend would be to change the `defaultHeight` in `RowRenderer` to 18. It's currently set to 16, but 18 is the true height of an unwrapped row. This will only help when wrap is off though! 😔

minnakt (approved):
let's try it out! :]

PR #201, opened 2023-02-15
EVG-18667 Add comment to log type enum

khelif96:
EVG-18867

### Description 
Adds a comment describing why this enum does not follow typical enum naming conventions.

### DIFF
--- a/src/constants/logURLTemplates.ts
+++ b/src/constants/logURLTemplates.ts
@@ -99,7 +99,7 @@ const getResmokeLogURL = (
   return `${logkeeperURL}/build/${buildID}/all?${stringifyQuery(params)}`;
 };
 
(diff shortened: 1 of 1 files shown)


src/constants/logURLTemplates.ts
```diff
@@ -99,6 +99,7 @@ const getResmokeLogURL = (
   return `${logkeeperURL}/build/${buildID}/all?${stringifyQuery(params)}`;
 };
 
+// Although this does not follow typical enum naming conventions, it is used to map the origin of the log to the type parameter in the Evergreen URL.
```
sophstad, line 102:
```suggestion
// Although this does not follow typical enum capitlization conventions, it is used to map the origin of the log to the type parameter in the Evergreen URL.
```

sophstad (approved):
Ty! 🥳

PR #206, opened 2023-02-22
EVG-18944: Upgrade button

sophstad:
EVG-18944

### Description 

- Changes in https://github.com/mongodb/leafygreen-ui/pull/1611 mean that we don't need to use wrappers around buttons that function as tooltip triggers. Previously the wrapper was required or else a disabled button would block tooltip rendering.
- The package changes also mean that we need to check for the `aria-disabled` attribute instead of just using Cypress/React Testing Library's "disabled" check. There are some details [here](https://github.com/mongodb/leafygreen-ui/pull/1611#discussion_r1099090020).

### Testing 

- Update tests


khelif96 (approved):
LGTM 💯

PR #227, opened 2023-03-17
EVG-19204 Prevent line numbers from being copied to the clipboard

khelif96:
EVG-19204

### Description 
When trying to copy multiple lines the line numbers would be copied onto the clipboard despite them not being selected.

### Testing 
Attempted to copy multiple lines in both chrome and safari and did not see line numbers logged.
Also wrote unit tests

### DIFF
--- a/src/components/LogRow/BaseRow/index.tsx
+++ b/src/components/LogRow/BaseRow/index.tsx
@@ -123,7 +123,7 @@ const BaseRow = forwardRef<any, BaseRowProps>((props, ref) => {
         onClick={handleClick}
         size="small"
       />
-      <Index>{lineNumber}</Index>
+      <Index lineNumber={lineNumber} />
       <StyledPre shouldWrap={wrap}>
         <ProcessedBaseRow
           color={resmokeRowColor}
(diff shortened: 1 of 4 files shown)


src/components/LogRow/BaseRow/index.tsx
```diff
@@ -235,6 +235,10 @@ const Index = styled.pre`
   line-height: inherit;
   font-size: inherit;
   user-select: none;
+
+  ::before {
+    ${({ lineNumber }) => lineNumber && `content: "${lineNumber}";`}
+  }
```
minnakt, line 241: The truthiness check is preventing `0` from being printed on the first line:
```suggestion
  ::before {
    ${({ lineNumber }) => `content: "${lineNumber}";`}
  }
```

Screenshot 2023-03-17 at 3 16 18 PM

minnakt (approved):
smart workaround! :] and confirmed I don't see it in Safari either!

src/components/LogRow/BaseRow/index.tsx
khelif96: Great catch!

PR #234, opened 2023-03-28
EVG-18101: Render task name and status

sophstad:
EVG-18101

### Description 

- Render task name and status on evergreen task log page. I copied over a lot of Spruce code for rendering the status badge.
- For test log pages, show the task status and additional breadcrumb indicating that a test log is currently being viewed.

### Screenshots

Task header:
image

Test header:
image

### Testing 

- Add integration test


src/components/TaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,108 @@
+import { css } from "@emotion/react";
```
khelif96, line 1: Can you write some stories for this?

src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -0,0 +1,49 @@
...
+      {!loading && (
+        <>
+          <StyledLink
+            data-cy="spruce-link"
+            href={taskLink}
+            onClick={() => sendEvent({ name: "Opened Task Link" })}
+            target="_blank"
+            title="Open task in Spruce"
+          >
+            {task?.displayName || "Task"}
```
khelif96, line 42: Could we include a more complete breadcrumb?
The designs feature something like this
image

image

image

khelif96 (changes requested):

src/vite-env.d.ts
```diff
@@ -1 +1,11 @@
 /// <reference types="vite/client" />
+
+declare module "*.gql" {
+  const content: any;
+  export default content;
+}
```
minnakt, line 6: Is it possible to type these like this?

```suggestion
declare module "*.gql" {
  import { DocumentNode } from "graphql";
  const content: DocumentNode;
  export default content;
}
```

cypress/integration/ansiiLogs/ansii_logView.ts
```diff
@@ -38,6 +38,17 @@ describe("Basic evergreen log view", () => {
     );
     cy.get(".ReactVirtualized__Grid").scrollTo("right");
   });
+  it("log header should show the log name and status and link to Spruce", () => {
```
minnakt, line 41: could we write a corresponding test for `resmoke_logView`?

src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const EvergreenTaskSubHeader: React.FC<Props> = ({
+  taskID,
+  execution,
+}) => {
+  const { sendEvent } = usePreferencesAnalytics();
+  const { data, loading } = useQuery<GetTaskQuery, GetTaskQueryVariables>(
+    GET_TASK,
+    {
+      variables: { taskId: taskID as string, execution: Number(execution) },
```
minnakt, line 23: do we need to cast the `taskID` and `execution` if they already come in as a `string` and `number` in the props?

src/components/TaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+  ${({ border }) => border && `border-color: ${border} !important;`}
+  ${({ fill }) => fill && `background-color: ${fill} !important;`}
+  ${({ text }) => text && `color: ${text} !important;`}
+  ${badgeWidthMaxContent}
+`;
+
+interface TaskStatusBadgeProps {
+  status: string;
+}
+const TaskStatusBadge: React.VFC<TaskStatusBadgeProps> = ({ status }) => {
```
minnakt, line 32:
```suggestion
const TaskStatusBadge: React.FC<TaskStatusBadgeProps> = ({ status }) => {
```

src/components/TaskStatusBadge/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return null;
+  }
+
+  let displayStatus = status;
+  if (taskStatusToCopy[status] === undefined) {
+    const err = new Error(`Status '${status}' is not a valid task status`);
+    reportError(err).warning();
+  } else {
+    displayStatus = taskStatusToCopy[status];
+  }
```
minnakt, line 43: Kind of wondering if we need this, though I know it's directly from Spruce

```suggestion
  const displayStatus = taskStatusToCopy[status] ?? status;
```

minnakt (changes requested):

src/components/TaskStatusBadge/index.tsx
sophstad: Since we're reporting the error in Spruce it seems reasonable to me to omit this logic in Parsley

src/components/SubHeader/EvergreenTaskSubHeader.tsx
sophstad: Fwiw I don’t feel that breadcrumbs make a ton of sense here. They’re useful for situating users in an information hierarchy but Parsley has no such hierarchy, Spruce does. Linking them directly to the task which then contains the breadcrumbs seems reasonable and less cluttered, but 🤷‍♀️

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+  text: string;
+  to?: string;
+  href?: string;
+  onClick?: () => void;
+  "data-cy"?: string;
+}
+interface BreadcrumbsProps {
+  breadcrumbs: Breadcrumb[];
+}
+const Breadcrumbs: React.VFC<BreadcrumbsProps> = ({ breadcrumbs }) => (
```
minnakt, line 22:
```suggestion
const Breadcrumbs: React.FC<BreadcrumbsProps> = ({ breadcrumbs }) => (
```

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        )}
+      </Fragment>
+    ))}
+  </Container>
+);
+
+interface BreadcrumbFragmentProps {
+  breadcrumb: Breadcrumb;
+}
+const BreadcrumbFragment: React.VFC<BreadcrumbFragmentProps> = ({
```
minnakt, line 43:
```suggestion
const BreadcrumbFragment: React.FC<BreadcrumbFragmentProps> = ({
```

src/components/TaskStatusBadge/TaskStatusBadge.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+      {Object.values(TaskStatus).map((status) => (
+        <TaskStatusBadge key={status} status={status} />
+      ))}
+    </Container>
+  ),
+};
+
+const Container = styled.div`
+  display: flex;
+  gap: 8px;
```
minnakt, line 35: nit. can use token

src/components/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { palette } from "@leafygreen-ui/palette";
+import Tooltip from "@leafygreen-ui/tooltip";
+import Icon from "components/Icon";
+import { StyledLink, StyledRouterLink } from "components/styles";
+import { size } from "constants/tokens";
+import { trimStringFromMiddle } from "utils/string";
+
+const { gray } = palette;
+
+export interface Breadcrumb {
```
minnakt, line 12: might also be worth copying the `Breadcrumbs` test from Spruce!

minnakt (changes requested):

cypress/integration/resmokeLogs/resmoke_logView.ts
```diff
@@ -34,6 +34,9 @@ describe("Basic resmoke log view", () => {
     );
     cy.get(".ReactVirtualized__Grid").scrollTo("right");
   });
+  it("log header should not show breadcrumbs since the task is not found by Evergreen", () => {
```
khelif96, line 37: Sorry I didn't clarify this earlier. You need to use the logkeeperBuildMetadata resolver to get this data. It can probably be updated to return a full task instead of just the taskId 
image

khelif96: I can put up a quick pr to surface this data. 
```graphql
{
  logkeeperBuildMetadata(buildId: "7e208050e166b1a9025c817b67eee48d") {
    task {
      displayName
      patchNumber
      status
      versionMetadata {
        id
        isPatch
        projectIdentifier
        message
        revision
      }
    }
  }
}
```

khelif96 (changes requested):

sophstad:
evergreen retry

cypress/integration/resmokeLogs/resmoke_logView.ts
sophstad: Done!

package.json
```diff
@@ -113,6 +113,7 @@
     "@typescript-eslint/parser": "5.48.0",
```
minnakt, line 113: could you do me a random favor and update the eslint:strict command in this file to `"eslint:strict": "STRICT=1 eslint '*.{js,ts,tsx}' 'src/**/*.ts?(x)' 'cypress/**/*.ts' 'src/gql/**/*.graphql'"`, I couldn't do it in my lint PR because it errors when there are no GraphQL files 🥹

minnakt (approved):
Great work Sophie!!! 🤩 I think you can close this random ticket ([EVG-6772](https://jira.mongodb.org/browse/EVG-6772)) that was floating in the epic too, what do you think?

src/components/styles/GlobalStyles.tsx
```diff
@@ -17,6 +17,7 @@ const globalStyles = css`
   body {
     font-family: "Euclid Circular A", "Helvetica Neue", Helvetica, Arial,
       sans-serif;
+    font-size: 13px;
```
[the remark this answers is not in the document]
sophstad: The provider already defaults to 13 but it only only applies to LeafyGreen elements anyway—it's helpful to set an accurate base size so that non-LG Typography elements (of which there are admittedly few!) are at the right scale.

src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+      "data-cy": "task-breadcrumb",
+      onClick: () => {
+        sendEvent({ name: "Opened Task Link" });
+      },
+    },
+    ...(testID
+      ? [
+          {
+            "data-cy": "test-breadcrumb",
+            tooltipText: currentTest?.name,
```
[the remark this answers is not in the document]
sophstad: I went with the tooltip approach because test names quickly got suuuuper long 😬 open to ideas but this seemed like a quick and straightforward approach!

src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  if (loading || logkeeperLoading || (!task && !logkeeperBuildMetadata)) {
+    return <Icon glyph="EvergreenLogo" size={24} />;
+  }
+
+  // @ts-expect-error - We verify above that either task or logkeeperBuildMetadata exists, so logkeeperBuildMetadata will not be undefined as TypeScript suspects
+  const loadedTask = task ?? logkeeperBuildMetadata.task;
+  const { displayName, patchNumber, status, versionMetadata } = loadedTask;
+  const { isPatch, projectIdentifier, message, revision } =
+    versionMetadata ?? {};
```
khelif96, line 57: You can avoid the ts errors by reorganizing the code
```ts
  const { task } = data ?? {};
  const { logkeeperBuildMetadata } = logkeeperData ?? {};
  const loadedTask = task ?? logkeeperBuildMetadata?.task;
  const { displayName, patchNumber, status, versionMetadata } =
    loadedTask || {};

  const { isPatch, projectIdentifier, message, revision } =
    versionMetadata || {};
```

khelif96 (changes requested):

src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+        `Patch ${patchNumber}`
+      ) : (
+        <InlineCode>{shortenGithash(revision)}</InlineCode>
+      ),
+    },
+    {
+      href: getEvergreenTaskURL(taskID, execution),
+      text: (
+        <>
+          {displayName} <TaskStatusBadge status={status} />
```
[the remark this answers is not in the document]
sophstad: I actually put it in its current location because it represents the task status, not test status. I think having it at the end could be misleading e.g. if the task succeeded but the test failed and there's a succeeded badge.

src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const { data, loading } = useQuery<TaskQuery, TaskQueryVariables>(GET_TASK, {
+    variables: { taskId: taskID, execution },
+    skip: logType === LogTypes.RESMOKE_LOGS,
+  });
+  const { data: logkeeperData, loading: logkeeperLoading } = useQuery<
+    LogkeeperTaskQuery,
+    LogkeeperTaskQueryVariables
+  >(GET_LOGKEEPER_TASK, {
+    variables: { buildId: buildID },
+    skip: logType !== LogTypes.RESMOKE_LOGS || !buildID,
```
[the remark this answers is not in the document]
sophstad: I actually think adding a toast if this query fails is too aggressive since users didn't take any action to invoke this query and wouldn't be able to act on the error. Since we [report GQL errors](https://github.com/evergreen-ci/parsley/blob/5c73d8fb269b1d2f693008553d02ef8903042856/src/gql/GQLProvider.tsx#L13) to Bugsnag I think our bases are covered.

src/components/SubHeader/EvergreenTaskSubHeader.tsx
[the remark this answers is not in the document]
minnakt: I feel like being able to quickly see the test name might be valuable when viewing different resmoke test logs in multiple tabs. One option could be to allow the test name specifically to have a higher character cut off than 30?

Screenshot 2023-04-07 at 10 38 33 AM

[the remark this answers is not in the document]
sophstad: done

src/components/SubHeader/EvergreenTaskSubHeader.tsx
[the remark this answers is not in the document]
khelif96: This would never be the case if the test fails then the task will also fail. So the statuses should always line up.

.eslintrc.js
```diff
@@ -213,6 +213,8 @@ module.exports = {
           ERROR,
           { selections: ["OperationDefinition", "FragmentDefinition"] },
         ],
+        // Following rule can possibly be removed after ESLint updates.
```
khelif96, line 216: For context was this rule disabled?

src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -45,16 +47,20 @@ export const EvergreenTaskSubHeader: React.FC<Props> = ({
 
   const { task } = data ?? {};
   const { logkeeperBuildMetadata } = logkeeperData ?? {};
+  const loadedTask = task ?? logkeeperBuildMetadata?.task;
 
-  if (loading || logkeeperLoading || (!task && !logkeeperBuildMetadata)) {
+  if (loading || logkeeperLoading || !loadedTask) {
     return <Icon glyph="EvergreenLogo" size={24} />;
```
khelif96, line 53: Can we show some placeholder text here in case we can't load the task?

src/components/SubHeader/EvergreenTaskSubHeader.tsx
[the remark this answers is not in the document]
khelif96: That specific scenario wouldn't ever happen because a test failure will always cause the task to fail. But the logic is sound for the opposite successful test but failed task. 👍

khelif96 (approved):
LGTM Great job!

.eslintrc.js
sophstad: We disable the same rule [in Spruce](https://github.com/evergreen-ci/spruce/blob/71b82e4f42f9bd4aaf790d22b1126e56cb92ed62/.eslintrc.js#L236-L237); when it's on fragment imports are interpreted as comments and the linter expects a space after the `#`.

src/components/SubHeader/EvergreenTaskSubHeader.tsx
sophstad: Good point we can still show the task link like we were doing previously 👍

PR #250, opened 2023-04-03
EVG-19318 Memoize analytics hooks

khelif96:
EVG-19318

### Description 
While working on #233 I noticed that components were rerendering due to `sendEvent` being initialized on each render. Added some memoization to the analytics hooks to prevent these rerenders. This eliminates one render 

Waiting on #249

### DIFF
--- a/src/analytics/loadingPage/useLogDownloadAnalytics.ts
+++ b/src/analytics/loadingPage/useLogDownloadAnalytics.ts
@@ -1,3 +1,4 @@
+import { useCallback, useMemo } from "react";
 import {
   Analytics as A,
   Properties,
@@ -13,11 +14,11 @@ interface P extends Properties {}
 interface Analytics extends A<Action> {}
 
 export const useLogDownloadAnalytics = (): Analytics => {
-  const sendEvent: Analytics["sendEvent"] = (action) => {
+  const sendEvent: Analytics["sendEvent"] = useCallback((action) => {
     addPageAction<Action, P>(action, {
       object: "LoadingPage",
     });
-  };
+  }, []);
 
-  return { sendEvent };
+  return useMemo(() => ({ sendEvent }), [sendEvent]);
 };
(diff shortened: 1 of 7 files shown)


src/analytics/shortcuts/useShortcutAnalytics.ts
```diff
@@ -1,3 +1,4 @@
+import { useCallback, useMemo } from "react";
```
minnakt, line 1: (comment placement is random) does `useLogDropAnalytics` need one too?

minnakt (approved):
nice detective work 🕵️🔎

src/analytics/loadingPage/useLogDownloadAnalytics.ts
```diff
@@ -13,11 +14,11 @@ interface P extends Properties {}
 interface Analytics extends A<Action> {}
 
 export const useLogDownloadAnalytics = (): Analytics => {
-  const sendEvent: Analytics["sendEvent"] = (action) => {
+  const sendEvent: Analytics["sendEvent"] = useCallback((action) => {
```
sophstad, line 17: These hooks are all so similar, do you think it would be worth creating a reusable base analytics hook that takes in the `object` string? Understandable if that seems like overkill though!

khelif96: What a great suggestion! Done. What do you think?

src/analytics/loadingPage/useLogDownloadAnalytics.ts
```diff
@@ -13,11 +14,11 @@ interface P extends Properties {}
 interface Analytics extends A<Action> {}
 
 export const useLogDownloadAnalytics = (): Analytics => {
```
sophstad, line 16: Looks great! I think we can use it here too
```suggestion
export const useLogDownloadAnalytics = () =>
   useAnalyticsRoot<Action>("LoadingPage");
```

sophstad (approved):

PR #254, opened 2023-04-05
EVG-18621: Increase error state visibility

sophstad:
EVG-18621

### Description 

Use LeafyGreen text input error state with tooltip on main search bar

### Screenshots

image

### DIFF
--- a/src/components/Search/SearchBar/index.tsx
+++ b/src/components/Search/SearchBar/index.tsx
@@ -1,21 +1,18 @@
 import { KeyboardEvent, useRef, useState } from "react";
 import styled from "@emotion/styled";
 import IconButton from "@leafygreen-ui/icon-button";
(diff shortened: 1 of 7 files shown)


src/components/Search/SearchBar/index.tsx
```diff
@@ -154,19 +152,20 @@ const SearchBar: React.FC<SearchBarProps> = ({
...
               {validatorMessage}
-            </IconWithTooltip>
+            </Tooltip>
           )
         }
         onChange={(e) => handleOnChange(e.target.value)}
         onKeyDown={handleKeyDown}
         placeholder="optional, regexp to search"
         spellCheck={false}
+        state={isValid ? "none" : "error"}
```
khelif96, line 168: I wonder if we can over ride this focus state and make it red too
image

khelif96 (approved):

PR #259, opened 2023-04-07
EVG-18852 Utilize streaming to download larger log files

khelif96:
EVG-18852

### Description 
Utilize a readable stream to download log files in chunks. This allows us to download log files and process them as they come in instead of trying to load them into memory. We can now load files of any size so long as they are under the 4gb memory limit browsers have. 

Example 2gb 2.2 million line log file:
image

### Testing 
Try to load any log file into parsley.
You can query Splunk for `index=logkeeper  log size stats num_lines > 1000000` to find really large log files.


src/hooks/useLogDownloader/utils.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!response.body) {
+    throw new Error("Network response has no body");
+  }
+
+  const reader = response.body.getReader();
+
+  return new ReadableStream({
+    async start(controller) {
+      try {
+        while (true) {
```
minnakt, line 36: I think you might want to tweak the ESLint settings to get rid of the warning

sophstad (approved):
how exciting 😊

jest.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     "^uuid$": "<rootDir>/node_modules/uuid/dist/index.js",
   },
   modulePaths: ["<rootDir>/src"],
-  setupFiles: ["whatwg-fetch"],
+  setupFiles: ["whatwg-fetch", "./jest.setup.ts"],
```
minnakt, line 13: if we're using a Jest setup file now, could we import `whatwg-fetch` there instead? The reason `whatwg-fetch` was on its own was because it avoided having a one line file with an import statement

minnakt (approved):

PR #260, opened 2023-04-07
EVG-18439 Implement updated loading page state

khelif96:
EVG-18439

### Description 
Building off of the download size logic in EVG-18852 this now shows the loading page with file size and new indeterminate loading bar.

Once #259 is merged I will rebase this commit and only include the relevant changes
### Screenshots

https://user-images.githubusercontent.com/4605522/230643572-40aff20c-582f-4677-ba89-8ac4b0168472.mov


src/pages/LogView/LoadingPage/LoadingBar/index.tsx
```diff
@@ -36,7 +39,31 @@ const Bar = styled.div<{ progress: number }>`
...
 `;
+
+const indeterminateAnimation = css`
+  position: relative;
+  bottom: 0;
+  top: 0;
+  width: 50%;
+
+  /* Move the bar infinitely */
+  animation: indeterminate-progress-bar 2s infinite;
```
minnakt, line 59: nit. probably not that important and not sure if it's fixable, but wanted to flag that the animation freezes before the log loads in (while the leaf is still moving properly)

https://user-images.githubusercontent.com/47064971/231023960-9c76b958-828c-4880-aacc-411fe88c442c.mov

minnakt (approved):
the animation looks great! 🥹

src/pages/LogView/LoadingPage/LoadingBar/index.tsx
khelif96: I can't figure out a way around it it freezes when processing the log which is an expensive operation. This should only happen for resmoke logs.

PR #261, opened 2023-04-07
EVG-18990 Override Home and End button functionality in Paginated view

khelif96:
EVG-18990

### Description 
Since we are using a paginated list for the log view using the Home and End button would not take you to the top and bottom of the list. This PR overrides the default behavior and uses the scrollToLine function to jump automatically to the first and last line even if the view is paginated. 
### Screenshots

Before:
https://user-images.githubusercontent.com/4605522/230675781-b80dfba5-03fd-437c-8422-ce9ddfe6ad29.mov

After:
https://user-images.githubusercontent.com/4605522/230675772-eec8340f-bfce-4f8e-86df-b9477109655a.mov

### Testing 
Lower paginationThreshold and use Home and End button to jump to parts of the log

### DIFF
--- a/src/components/PaginatedVirtualList/index.tsx
+++ b/src/components/PaginatedVirtualList/index.tsx
@@ -1,5 +1,7 @@
 import { forwardRef, useCallback, useEffect, useRef } from "react";
 import { ItemContent, Virtuoso, VirtuosoHandle } from "react-virtuoso";
+import { CharKey } from "constants/keys";
+import { useKeyboardShortcut } from "hooks";
 import { PaginatedVirtualListRef } from "./types";
 import usePaginatedVirtualList from "./usePaginatedVirtualList";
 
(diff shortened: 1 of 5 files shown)


src/components/PaginatedVirtualList/index.tsx
```diff
@@ -50,6 +52,14 @@ const PaginatedVirtualList = forwardRef<
       ref: listRef,
     });
 
+    useKeyboardShortcut({ charKey: CharKey.PageEnd }, () => {
```
minnakt, line 55: Would it be worthwhile to add end-to-end tests for these?

minnakt (approved):

src/components/PaginatedVirtualList/index.tsx
khelif96: Since we never actually invoke the paginated list in our e2e tests.

PR #262, opened 2023-04-11
EVG-19266: Log username with BugSnag errors

sophstad:
EVG-19266

### Description 

Fetch user ID and include it in BugSnag error reporting


src/components/ErrorBoundary/index.tsx
```diff
@@ -93,6 +93,8 @@ const ErrorBoundary: React.FC<{ children: React.ReactNode }> = ({
   const ErrorBoundaryComp = getBoundary();
 
   const onError = (event: Event) => {
+    const userId = localStorage.getItem("userId") ?? undefined;
```
khelif96, line 96: Would love to see this added in spruce as well ❤️

khelif96 (approved):

PR #265, opened 2023-04-17
EVG-19623 Limit streamed log file download size

khelif96:
EVG-19623

### Description 
* Limits the maximum file size to 2.5 GB
* Ensures we still show a log if the server cancels the request

### Screenshots

https://user-images.githubusercontent.com/4605522/232575596-f79451d3-36a3-4b33-8277-a20f540d7a90.mov

### Testing 
Update the `downloadSizeLimit` to be a size small enough that you can test against local data
Set the browser tab to offline to mimic the request being closed by the load balancer.

### DIFF
--- a/src/hooks/useLogDownloader/index.ts
+++ b/src/hooks/useLogDownloader/index.ts
@@ -1,6 +1,7 @@
 import { useEffect, useState } from "react";
 import { useLogDownloadAnalytics } from "analytics";
 import { LogTypes } from "constants/enums";
+import { LOG_FILE_SIZE_LIMIT } from "constants/logs";
 import useStateRef from "hooks/useStateRef";
 import { isProduction } from "utils/environmentVariables";
 import { leaveBreadcrumb, reportError } from "utils/errorReporting";
@@ -17,7 +18,11 @@ import { getBytesAsString } from "utils/string";
  * - error: an error message if the download fails
  * - fileSize: the size of the log file in bytes
  */
-const useLogDownloader = (url: string, logType: LogTypes) => {
+const useLogDownloader = (
+  url: string,
+  logType: LogTypes,
+  downloadSizeLimit: number = LOG_FILE_SIZE_LIMIT
+) => {
   const [data, setData] = useState<string[] | undefined>();
   const [error, setError] = useState<string | undefined>();
   const [fileSize, setFileSize, getFileSize] = useStateRef<number>(0);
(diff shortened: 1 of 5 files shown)


src/hooks/useLogDownloader/index.ts
```diff
@@ -34,6 +42,21 @@ const useLogDownloader = (url: string, logType: LogTypes) => {
...
       },
+      downloadSizeLimit,
+      onIncompleteDownload: (reason, incompleteDownloadError) => {
+        leaveBreadcrumb(
+          "useLogDownloader",
+          { incompleteDownloadError, reason },
+          "error"
+        );
+        dispatchToast.warning(
+          "We were only able to partially download this log. Use the Evergreen CLI command in the details menu to download the log on to your machine.",
```
sophstad, line 53:
```suggestion
          "Parsley was only able to partially download this log. Use the Evergreen CLI command in the details menu to download the log onto your machine.",
```

src/utils/fetchLogFile/index.ts
```diff
@@ -61,7 +65,15 @@ const streamedFetch = async (url: string, options: StreamedFetchOptions) => {
           controller.enqueue(value);
         }
       } catch (error) {
-        controller.error(error);
+        // If we hit an error, but we've already fetched some bytes, then we can assume that
+        // the download was incomplete. This is because the load balancer will close the connection if
```
sophstad, line 69:
```suggestion
        // the download was incomplete. This is because the controller will close the connection if
```

src/utils/fetchLogFile/index.ts
```diff
@@ -1,7 +1,11 @@
+type IncompleteDownloadReason = "SERVER_ERROR" | "FILE_TOO_LARGE";
```
sophstad, line 1: Any reason not to use a string enum here?

khelif96: No, it was a preference but I updated it to use an enum.

src/utils/fetchLogFile/index.ts
```diff
@@ -1,7 +1,15 @@
+enum IncompleteDownloadReason {
+  SERVER_ERROR = "SERVER_ERROR",
+  FILE_TOO_LARGE = "FILE_TOO_LARGE",
```
sophstad, line 3: [nit] Standard uppercase would be more in line with enums across Parsley
```suggestion
  ServerError = "SERVER_ERROR",
  FileTooLarge = "FILE_TOO_LARGE",
```

sophstad (approved):

PR #267, opened 2023-04-18
EVG-18854 Add instructions for downloading logs via the CLI 

khelif96:
EVG-18854

### Description 
This builds off of #265 and adds download instructions for logs.
Only commits that are relevant to this pr are da417569fd4989d23a2bc1b3d63819bf14ecd944 and onwards 
### Screenshots
image


src/components/DetailsMenu/index.tsx
```diff
@@ -29,6 +30,11 @@ const DetailsMenu: React.FC<DetailsMenuProps> = ({ "data-cy": dataCy }) => (
       </Column>
     </Row>
     <ButtonRow />
+    <Row>
+      <div>
+        <CLIInstructions />
+      </div>
+    </Row>
   </DetailsMenuCard>
```
minnakt, line 38: (comment placement is random) Do you want to update the JIRA button text as shown in the designs? We had talked about changing the copy during design sync because the button text makes it sound like it's a link to JIRA

minnakt (changes requested):

src/hooks/useLogDownloader/index.ts
```diff
@@ -34,6 +42,21 @@ const useLogDownloader = (url: string, logType: LogTypes) => {
...
       },
+      downloadSizeLimit,
+      onIncompleteDownload: (reason, incompleteDownloadError) => {
+        leaveBreadcrumb(
+          "useLogDownloader",
+          { incompleteDownloadError, reason },
+          "error"
+        );
+        dispatchToast.warning(
+          "We were only able to partially download this log. Use the Evergreen CLI command in the details menu to download the log on to your machine.",
```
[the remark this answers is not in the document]
khelif96: This isn't included as part of the diff for this pr. (See description) But this statement would not be true since local downloads are also susceptible to timeouts.

src/components/DetailsMenu/CLIInstructions/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+    case LogTypes.EVERGREEN_TEST_LOGS:
+      if (!taskID || !execution || !testID) {
+        return "";
+      }
+      return `evergreen buildlogger fetch --task_id ${taskID} --execution ${execution} --test_name ${testID}`;
+    case LogTypes.RESMOKE_LOGS:
+      if (!buildID) {
+        return "";
+      }
+      return `curl ${getResmokeLogURL(buildID || "", { testID, raw: true })}`;
```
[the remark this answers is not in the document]
khelif96: Thats odd I don't get that
image

src/components/DetailsMenu/index.tsx
khelif96: image
Done

src/components/DetailsMenu/CLIInstructions/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const getCLICommand = (logMetadata: LogMetadata) => {
+  const { taskID, execution, testID, logType, origin, buildID } = logMetadata;
+  switch (logType) {
+    case LogTypes.EVERGREEN_TASK_LOGS:
+      if (!taskID || !execution || !origin) {
+        return "";
+      }
+      return `evergreen buildlogger fetch --task_id ${taskID} --execution ${execution} --tags ${origin}_log`;
```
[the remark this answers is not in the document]
khelif96: Great catch! omitting the param does the same thing so I did that instead

src/context/auth/auth.test.tsx
```diff
@@ -48,18 +48,18 @@ describe("auth", () => {
...
-    expect(() => render(<Component />)).toThrow(
+    const { result } = renderHook(() => useAuthContext());
+
+    expect(result.error?.message).toBe(
       "useAuthContext must be used within an AuthProvider"
     );
   });
 
-  it("should execute a query against GraphQL upon mount to check if user is authenticated", () => {
+  it("should execute a query against GraphQL upon mount to check if user is authenticated", async () => {
```
minnakt, line 58: unused `async` maybe?

src/context/auth/auth.test.tsx
```diff
@@ -131,33 +125,29 @@ describe("auth", () => {
...
       jest.spyOn(global, "fetch").mockImplementation(mockFetchPromise);
 
-      const { Component, hook } = renderComponentWithHook();
-      render(<Component />, { wrapper });
-
-      act(() => {
-        hook.current.logoutAndRedirect();
+      renderHook(() => useAuthContext(), {
+        wrapper,
       });
```
minnakt, line 146: I think this test should still be calling `hook.current.logoutAndRedirect`, since the test block is testing this method

src/components/DetailsMenu/CLIInstructions/CliInstructions.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+        testID: "testId",
+        logType: LogTypes.EVERGREEN_TEST_LOGS,
+      });
+    });
+    const testDownloadCommand =
+      "evergreen buildlogger fetch --task_id taskId --execution 0 --test_name testId";
+    expect(screen.getByText(testDownloadCommand)).toBeInTheDocument();
+  });
+  it("should render the logkeeper download command if the log is a resmoke log", () => {
+    mockEnv("REACT_APP_LOGKEEPER_URL", "https://logkeeper-test.org");
```
minnakt, line 87: I think this doesn't actually do anything; the URL seems to still be `undefined` if you print `logkeeperDownloadCommand`

src/components/DetailsMenu/CLIInstructions/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+    case LogTypes.EVERGREEN_TEST_LOGS:
+      if (!taskID || !execution || !testID) {
+        return "";
+      }
+      return `evergreen buildlogger fetch --task_id ${taskID} --execution ${execution} --test_name ${testID}`;
+    case LogTypes.RESMOKE_LOGS:
+      if (!buildID) {
+        return "";
+      }
+      return `curl ${getResmokeLogURL(buildID, { testID, raw: true })}`;
```
minnakt, line 36: hmm this still doesn't work for me without quotes in zsh (I ran the command you showed in the other comment!)

minnakt (approved):
great test refactors!

src/context/auth/auth.test.tsx
khelif96: `logOutAndRedirects` gets called automatically by the `useEffect`. Which is what causes the component to be unmounted on render.

src/components/DetailsMenu/CLIInstructions/index.tsx
khelif96: Added quotes natively. Turns out they were being escaped by my terminal which caused it to work for me.

src/context/auth/auth.test.tsx
khelif96: The behavior of the component is to automatically redirect you to the login page. `logoutAndRedirect` was never actually causing this component to redirect it was always the redirect from the `useEffect`. If you remove this line from the original test it will still pass.

PR #270, opened 2023-04-24
EVG-19288: Show test status for resmoke logs

minnakt:
EVG-19288

### Description 
This PR shows the test status for resmoke test logs.

Note that for resmoke logs, task test IDs are different from logkeeper test IDs (which is intentional and not an implementation mistake). I asked Julian if we should expose the test status via Logkeeper, but he said that it would not be straightforward to implement, and that Logkeeper is going away in his upcoming project anyways. To match the task test IDs to the Logkeeper test IDs, I just went with a naive solution of matching on the associated log URL.

#### Other changes
I realized Jest wasn't properly interpreting GraphQL files which was preventing me from writing tests with Apollo's `MockedProvider`. I installed [`@graphql-tools/jest-transform`](https://the-guild.dev/graphql/tools/docs/api/modules/jest_transform_src) to transform the GraphQL files for us. I'm guessing the Babel config makes it so that we don't have to do this in Spruce.

### Screenshots

Screenshot 2023-04-24 at 10 11 53 AM

### Testing 
- Added stories for `TestStatusBadge`
- Added `EvergreenTaskSubHeader.test.tsx`


src/gql/fragments/base-task.graphql
```diff
@@ -4,6 +4,16 @@ fragment BaseTask on Task {
   id
   patchNumber
   status
+  tests(opts: null) {
```
khelif96, line 7: This would try to fetch test results for task logs where they aren't really relevant. This should only be requested when we fetch test logs.

src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -70,12 +71,15 @@ export const EvergreenTaskSubHeader: React.FC<Props> = ({
...
     status,
     versionMetadata,
+    tests,
   } = loadedTask;
   const { isPatch, projectIdentifier, message, revision } = versionMetadata;
 
-  const currentTest = logkeeperBuildMetadata?.tests?.find(
-    (test) => test.id === testID
-  );
+  const currentTest = isResmoke
```
khelif96, line 78: Is there a plan to do this for normal test results as well?

khelif96 (changes requested):
Don't forget to merge main.

src/components/SubHeader/EvergreenTaskSubHeader.tsx
minnakt: I think there's a problem with the format for normal test results. For example, if you look at the test results for [this task](https://spruce.mongodb.com/task/evergreen_ubuntu2004_test_thirdparty_docker_patch_4b2374178b940788892dddae673a2c103cfd21ca_6447ef851e2d17c4491fa5bd_23_04_25_15_19_34/tests?execution=0&sortBy=STATUS&sortDir=ASC), all of the tests have the same name `"output.thirdparty-docker.test"` in Parsley and are only distinguished by their values for `shareLine`. So it makes it kind of makes it difficult to tell what test the user actually clicked on initially

minnakt: hello I have looked 🧐 I feel like it's technically possible since `parsleyUrl` does store the `shareLine` for each test but it's kind of handwavy since you would be matching on a number rather than a real ID. There is also question of how necessary the status is, since for Evergreen test logs the status is usually immediately visible versus resmoke test logs it's hard to tell at a glance

khelif96 (approved):
Great job!

PR #272, opened 2023-04-28
EVG-19733 Increase pagination offset to 200 to fix page jitter

khelif96:
EVG-19733

### Description 
A user reported seeing the log pane jitter when using the cmd + left and right buttons to paginate. We determined this was due to them having a large monitor which was rendering more lines than the pagination offset was set to. This lead to the virtual list jumping back and forth between each log pane.

Video from bug report:
https://user-images.githubusercontent.com/4605522/235185266-7e931621-3448-4451-96f9-b56219ccccfe.mov


minnakt (approved):
sounds good to me

khelif96:
evergreen merge

PR #281, opened 2023-05-03
EVG-19008: Implement collapsed row redesign

minnakt:
EVG-19008

### Description 
This PR updates the collapsed row according to Lara's designs in Figma. 

Notable changes:
* Text style is `Body`, not`Overline`
* Expand icon is vertical, not diagonal
* Row is thicker. Kind of nervous about this one! 🧍

### Screenshots
Screenshot 2023-05-03 at 2 11 02 PM

### Testing 
- Tests should pass

### DIFF
--- a/src/components/LogRow/CollapsedRow/index.tsx
+++ b/src/components/LogRow/CollapsedRow/index.tsx
@@ -85,7 +85,7 @@ CollapsedRow.displayName = "CollapsedRow";
 const CollapsedLineWrapper = styled.div`
   display: flex;
   align-items: center;
-  background-color: #f4f5f5;
+  background-color: #f4f5f5; // Custom gray background color.
   padding: 2px 0;
   padding-left: ${size.l};
 `;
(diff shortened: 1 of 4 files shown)


src/components/LogRow/CollapsedRow/index.tsx
```diff
@@ -63,21 +59,23 @@ const CollapsedRow: React.FC<CollapsedRowProps> = ({
...
-        All
-      </StyledButton>
-      <StyledButton
-        leftGlyph={<Icon glyph="Expand" />}
-        onClick={expandFive}
-        size="xsmall"
-      >
-        {SKIP_NUMBER} Above and Below
-      </StyledButton>
+      <StyledBody>{lineText}</StyledBody>
```
khelif96, line 62: Looks Great! I wonder if it would be possible to get the background to extend the full width but since this has always been there no need to fix it in this pr.
image

khelif96 (approved):

src/components/LogRow/CollapsedRow/index.tsx
minnakt: I'll take a look and see! 😅

PR #286, opened 2023-05-23
EVG-18703, EVG-19910 Ensure query params are properly re-encoded  when writing to query params

khelif96:
EVG-18703
EVG-19910

### Description 
Filtering and highlights perform a special form of uriEncoding when writing query parameters to ensure that they do not become corrupted or misinterpreted. Whenever a query parameter is written we need to rewrite all query parameters. Usually rewriting the query parameters would automatically reencode them but it appears this was not being applied to query parameters that were arrays.

### Screenshots

https://github.com/evergreen-ci/parsley/assets/4605522/e0f7227e-ac4b-4a84-afaf-e4b5356a79cc

### Testing 
Unit tests, testing against the files in the bug report


src/hooks/useHighlightParam/index.ts
sophstad, line 1: Now that we perform encoding in `useQueryParam`, is the specialized `useHighlightParam` still necessary?

khelif96: Yes since the special encoding within `useHighlightParam` serves the same purpose as the one in `useFilterParam`. They serve as an additional layer to ensure complex query and filter params are preserved in the URL since they have the potential to be broken when the URL is parsed.
You can test it yourself with the below steps. If you remove the encoding logic from `useHighlightParam` and apply a highlight with a comma you will see the highlights become broken up.
1. Remove encoding logic from `useHighlightParam`
2. Add a highlight filter that has a comma `some,example`
3. The added highlight will be broken up because the `,` is not properly parsed. 

I added a test to `useHighlightParam which tests this.

src/hooks/useQueryParam/index.ts
```diff
@@ -41,8 +41,21 @@ const useQueryParam = <T>(
```
sophstad, line 41: It would be nice to type the `param` parameter as `QueryParams` (maybe a followup ticket?)

sophstad (approved):

src/hooks/useQueryParam/index.ts
khelif96: Done!

PR #306, opened 2023-06-02
EVG-18922: Upgrade typography

sophstad:
EVG-18922

### Description 

- Upgrade typography package and replace custom typing with exported props


src/types/leafygreen.ts
```diff
@@ -1,26 +0,0 @@
-import Card from "@leafygreen-ui/card";
```
minnakt, line 1: (placement of comment is random) Does `BaseToggle/index.tsx` need `DisclaimerProps`?

src/types/leafygreen.ts
```diff
@@ -1,26 +0,0 @@
...
-import { H1, H2, H3, Overline, Subtitle } from "@leafygreen-ui/typography";
```
minnakt, line 2: (placement of comment is random) Does `SearchBarGuideCue/index.tsx` need `InlineKeyCodeProps`?

src/types/leafygreen.ts
sophstad: I mentioned [here](https://github.com/evergreen-ci/spruce/pull/1852#discussion_r1211764099) that I think it's sufficient to only add types when an error is thrown since that's we've done up until this point. These components don't throw errors, I think because they do not use LG's Polymorphic under the hood. Same applies below!

minnakt (approved):

PR #307, opened 2023-06-05
EVG-20080 introduce mockVirtuosoContext to parsley tests

khelif96:
EVG-20080

### Description 
I was reading through some of react virtuoso docs and came across [this section](https://virtuoso.dev/mocking-in-tests/) about mocking react virtuoso in tests. Since it normally does not work in the jest dom this change would help us improve our test coverage so I decided to quickly add it in.

### Testing 
This improves test coverage.


minnakt (approved):
cool find :)

PR #310, opened 2023-06-06
EVG-19895: Surface default filters in modal

minnakt:
EVG-19895

### Description 
I decided to split this ticket into two PRs because it was getting too big. This PR only represents the visual changes of showing default filters in the modal. The next PR will include the functionality of actually being able to interact with and apply them.

Also includes:
- Fix some z-index issues to prevent conflicts between modal and sidenav.

### Screenshots
Screenshot 2023-06-05 at 8 12 37 PM

### Testing 
- Added Storybook stories
- Jest and Cypress tests will be added in the next PR


src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -30,13 +32,25 @@ export const EvergreenTaskSubHeader: React.FC<Props> = ({
...
+      projectIdentifier: task?.versionMetadata?.projectIdentifier,
+    });
+  };
+
   const { data, loading } = useQuery<TaskQuery, TaskQueryVariables>(GET_TASK, {
     variables: { taskId: taskID, execution },
     skip: isResmoke,
+    onCompleted: ({ task }) => {
+      onQueryCompleted(task);
+    },
```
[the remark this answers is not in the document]
minnakt: thank you for the tip 🥹 I gave it a try!

src/constants/tokens.ts
```diff
@@ -13,8 +13,11 @@ const size = {
 
 const zIndex = {
   backdrop: -1,
-  drawer: 10,
-  modal: 20,
+
+  // Set these values to 1 to utilize LeafyGreen's built-in stacking context.
```
khelif96, line 17: Is this a new change? (please share I must have missed the announcement)

src/hooks/useCachedTask/index.ts
```diff
@@ -0,0 +1,24 @@
...
+      query: GET_LOGKEEPER_TASK,
+      variables: { buildId: buildID },
+    });
+    return data?.logkeeperBuildMetadata?.task;
+  }
+
+  const data = cache.readQuery<TaskQuery>({
+    query: GET_TASK,
+    variables: { taskId: taskID, execution: Number(execution) },
+  });
```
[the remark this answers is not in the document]
khelif96: If you use useQuery you can probably also eliminate some of the complexity in the hook or the hook as a whole since the logic should be fairly straightforward.

src/components/ApplyFiltersModal/ApplyFiltersModal.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+
+const Component = ({ ...args }) => {
+  const [, setSearchParams] = useQueryParams();
+  const { setLogMetadata } = useLogContext();
+  const [open, setOpen] = useState(false);
+
+  useEffect(() => {
+    cache.reset();
+    cache.writeQuery<TaskQuery, TaskQueryVariables>({
```
khelif96, line 26: This can be replaced with a simple mock apollo context if you use the useQuery method below.

src/components/ApplyFiltersModal/DefaultFilter.tsx
```diff
@@ -0,0 +1,76 @@
...
+                triggerEvent="hover"
+              >
+                This filter is already active.
+              </Tooltip>
+            )}
+          </>
+        }
+        name={filter.expression}
+      />
+      <BadgeContainer>
```
khelif96, line 46: Sorry if I missed the designs but have you considered reusing the existing filter components? It would be nice to have some consistency between them and this component. 


image

src/components/ApplyFiltersModal/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      css={css`
+        z-index: ${zIndex.modal};
+      `}
+      onCancel={() => setOpen(false)}
+      onConfirm={() => setOpen(false)}
+      open={open}
+      setOpen={setOpen}
+      title="Default Filters"
+    >
+      {parsleyFilters?.map((f) => (
```
khelif96, line 52: I think this would benefit from a scrollable area. (I'm not really sure what the Leafygreen convention is for overflowing modals) But I would imagine this section would grow pretty quickly as teams adopt this feature. (It will surely be really popular)

src/components/SidePanel/NavGroup/BaseNavGroup/index.tsx
```diff
@@ -25,6 +26,7 @@ const BaseNavGroup = <T extends {}>({
   items,
   navGroupTitle,
   defaultMessage,
+  additionalHeaderText,
```
khelif96, line 29: I feel like this prop is misleading since its a button that opens a modal but I can't think of an alternative name. 😅

khelif96 (changes requested):
Great job figuring out the cache! There is actually an easier way to use it that I had in mind sorry for not clarifying it earlier. Also had some suggestions for the interface.

src/hooks/useCachedTask/index.ts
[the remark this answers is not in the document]
minnakt: ohh my bad HAHA yeap I can update this

src/constants/tokens.ts
minnakt: Yep, I mentioned this change in the PR description! Spruce already has these same values so they're just copied over from there 🤠

src/components/SidePanel/NavGroup/BaseNavGroup/index.tsx
minnakt: Technically you could pass any kind of element here, since the modal logic is all external and not tied to this component

src/hooks/useCachedTask/index.ts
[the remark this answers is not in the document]
minnakt: Okay I tried it again, but this time I refactored some of our query logic for tasks. I'm thinking this should be okay given the default fetch policy of `cache-first`, so if it already exists in the cache we won't make a new network request (I checked the Network tab to verify this was true)

src/components/ApplyFiltersModal/DefaultFilter.tsx
minnakt: I'm not entirely sure about this pattern as I feel like it's somewhat of a waste of space to show something that will never be interactive for any of these filters. TBH I'm not even sure I want to show the case/match values in the first place, because I feel like most users will be like “oh yay, helpful filters someone else defined for me” and ignore everything but the string value anyways (and they can always just check the properties manually in the side panel). But I also think this would be super easy to change in the future with little detriment to user experience =]

src/components/SidePanel/NavGroup/FilterNavGroup/index.tsx
```diff
@@ -57,26 +65,49 @@ const FilterNavGroup: React.FC<FilterNavGroupProps> = ({
...
-        </FilterWrapper>
-      ))}
-    </BaseNavGroup>
+    <>
+      <ApplyFiltersModal open={open} setOpen={setOpen} />
+      <BaseNavGroup
+        additionalHeaderText={
+          // TODO: Unhide in EVG-19897.
+          isProduction ? null : (
+            <ModalTrigger onClick={() => setOpen(true)}>
```
khelif96, line 74: Since this span is now a button I think it would benefit from some accessibility enhancements. Normally this would have been caught by our linters but styled components prevent that.  
https://www.w3.org/WAI/ARIA/apg/patterns/button/examples/button/

khelif96 (approved):
Great Job on this! One final comment but approving to unblock.

src/components/ApplyFiltersModal/DefaultFilter.tsx
minnakt: good point, we can also change the design in that ticket if needed!!!

PR #313, opened 2023-06-13
EVG-19895: Enable interaction with project filters

minnakt:
EVG-19895

### Description 
This PR allows you to interact with and apply the default filters. 

(I think I'll open a separate ticket _outside_ of the epic for changing the `name` field to `expression` field.)

### Screenshots

https://github.com/evergreen-ci/parsley/assets/47064971/2e334c64-6404-4908-bdba-66cc4c785caa

### Testing 
- Added `modal_filters.ts` (will not pass until corresponding Evergreen PR is merged). In the future, we can also test saved personal filters in this file.
- Added `ApplyFiltersModal.test.tsx`

### Evergreen PR 
- evergreen-ci/evergreen/pull/6650


src/components/ApplyFiltersModal/DefaultFilter.tsx
```diff
@@ -3,32 +3,53 @@ import Badge from "@leafygreen-ui/badge";
...
+    }
+  };
 
   return (
     <DefaultFilterContainer data-cy="default-filter">
       <Checkbox
-        checked={alreadyActive}
-        disabled={alreadyActive}
+        checked={active || selected}
+        disabled={active}
```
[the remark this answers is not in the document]
minnakt: I'm not sure that I can exactly put it into words but something about removing filters through the modal seems like an odd UX! 

The unchecking mechanism makes a lot more sense to me when it's a real-time filter pop-up, like the ones we have in Spruce:

Screenshot 2023-06-13 at 9 17 44 PM


But for some reason, unchecking to remove strikes me as strange in this modal. 🤔 I guess my thinking is that there's a certain kind of finality to clicking that "Apply" button which makes me feel that the user is now the owner of that copy and responsible for managing it in the panel. 

Let me know what you think, it's really open-ended 😅 I'm not even 100% sure why it seems strange to me, it's just a feeling I'm having

src/components/ApplyFiltersModal/index.tsx
```diff
@@ -42,27 +48,60 @@ const ApplyFiltersModal: React.FC<ApplyFiltersModalProps> = ({
...
       setOpen={setOpen}
+      submitDisabled={state.selectedFilters.length === 0}
       title="Default Filters"
     >
       <Scrollable>
         {parsleyFilters?.map((f) => (
           <DefaultFilter
             key={f.expression}
             activeFilters={filters}
+            addFilter={(filterToAdd: Filter) =>
```
khelif96, line 94:
```suggestion
            addFilter={(filterToAdd) =>
```

nit. This should already be typed from the prop

src/components/ApplyFiltersModal/DefaultFilter.tsx
```diff
@@ -3,32 +3,53 @@ import Badge from "@leafygreen-ui/badge";
...
+  removeFilter,
   activeFilters,
+  selectedFilters,
   filter,
 }) => {
-  const alreadyActive = !!activeFilters.find(
-    (f) => f.name === filter.expression
-  );
+  const active = !!activeFilters.find((f) => f.name === filter.expression);
+  const selected = !!selectedFilters.find((f) => f.name === filter.expression);
```
khelif96, line 27: It seems odd that these values are calculated within the component instead of being passed in especially since they aren't being used for anything else.
Since this component is essentially a checkbox I don't think it should be responsible for determining its own state.

src/components/ApplyFiltersModal/index.tsx
```diff
@@ -42,27 +48,60 @@ const ApplyFiltersModal: React.FC<ApplyFiltersModalProps> = ({
...
       `}
-      onCancel={() => setOpen(false)}
-      onConfirm={() => setOpen(false)}
+      data-cy="apply-filters-modal"
+      onCancel={onCancel}
+      onConfirm={onConfirm}
       open={open}
       setOpen={setOpen}
+      submitDisabled={state.selectedFilters.length === 0}
       title="Default Filters"
```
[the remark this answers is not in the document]
minnakt: that makes sense, Bynn mentioned that he would prefer it to say "Project Filters" so I think I'm gonna go with that to keep the wordiness down

src/components/ApplyFiltersModal/state.ts
```diff
@@ -0,0 +1,46 @@
...
+  | { type: "REMOVE_FILTER"; filterToRemove: string }
+  | { type: "RESET" };
+
+type State = {
+  selectedFilters: Filter[];
+};
+
+const initialState = (): State => ({
+  selectedFilters: [],
+});
```
sophstad, line 15: I think a plain object suffices here
```suggestion
const initialState: State = {
  selectedFilters: [],
};
```

src/components/ApplyFiltersModal/DefaultFilter.tsx
sophstad: +1

minnakt:
evergreen retry

src/components/ApplyFiltersModal/state.ts
sophstad: That's really interesting, seems like a cool way to get around this is actually this [optional 3rd argument](https://react.dev/reference/react/useReducer#parameters) known as the initializer function! I thought [this writeup](https://mtm.dev/reset-usereducer-state) was helpful.

minnakt: thanks for explaining, I didn't know this 🙏

PR #315, opened 2023-06-16
EVG-19893: Offer collapsed view for filters

minnakt:
EVG-19893

### Description 
This PR enables collapsing filters in the side panel. The `Accordion` component is brought over from Spruce, but I removed some props that seem to have outlived their usefulness (e.g. `showCaret` and `toggleFromBottom` aren't used in Spruce at all) and tried to improve the animation by borrowing some techniques from LeafyGreen.

### About the Design
I asked a few users to pick between two designs: 
Option A     |  Option B
:-------------------------:|:-------------------------:
 Screenshot 2023-06-16 at 10 45 20 AM |  Screenshot 2023-06-16 at 10 45 05 AM

The users settled on Option A. They explained that they would only collapse a filter if they knew they weren't going to change any of its properties, and so they didn't need to have the context about case sensitivity or match type.

Open to more design suggestions if anyone has them 🙏

### Screenshots

https://github.com/evergreen-ci/parsley/assets/47064971/0de8f279-e53d-4457-839f-361431eccca7

### Testing 
- Added `Accordion.test.tsx`. It's difficult to test the collapsing/expanding for filters themselves because `toBeVisible` assertions don't work with `overflow`.


minnakt:
the PR is not actually that big, it's mostly storyshot updates ;-;

src/components/Accordion/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }, [children, childrenRef]);
+
+  const toggleAccordionHandler = () => {
+    setIsAccordionDisplayed(!isAccordionDisplayed);
+    onToggle({ isVisible: !isAccordionDisplayed });
+  };
+
+  return (
+    <div className={className} data-cy={dataCy}>
```
sophstad, line 52: What do you think about using `<details>` and `<summary>` to introduce some more semantic HTML? ((docs)[https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details])
```suggestion
    <details className={className} data-cy={dataCy}>
```

src/components/SidePanel/NavGroup/FilterNavGroup/FilterGroup/index.tsx
```diff
@@ -65,154 +69,175 @@ const FilterGroup: React.FC<FilterGroupProps> = ({
       ? "Filter cannot be empty"
       : `Invalid Regular Expression: ${getRegexpError(newFilterName)}`;
 
+  const iconTooltip =
+    !isValid && !isEditing ? (
+      <IconWithTooltip color={red.base} glyph="ImportantWithCircle">
+        Invalid filter expression, please update it!
+        <Error>{validationMessage}</Error>
+      </IconWithTooltip>
+    ) : null;
```
sophstad, line 78: [super nit] My preference would be to put the conditionals outside of the `iconTooltip` var itself

sophstad (commented):
Works really well! I prefer the design that the users picked too.

src/components/Accordion/index.tsx
minnakt: I wanted this to work so badly  but then I ran into a problem where I couldn't get it to animate nicely 😔 The issue is also described in the docs:

Screenshot 2023-06-16 at 2 52 22 PM

I think it's not too bad to keep it as a div since we have `role="button"` and `aria-expanded` attributes, but let me know if any other accessibility attributes come to mind that I should add!

khelif96: That's a shame, also cool find @sophstad! didn't know this existed.

src/components/Accordion/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <AccordionToggle
+        data-cy="accordion-toggle"
+        onClick={toggleAccordionHandler}
+        role="button"
+      >
+        <AccordionIcon
```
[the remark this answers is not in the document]
minnakt: I went with `gray.dark1` so that it's the same color as the badge!

src/components/styles/Typography.tsx
```diff
@@ -12,3 +12,9 @@ export const wordBreakCss = css`
 export const WordBreak = styled.span`
   ${wordBreakCss};
 `;
+
+export const TextEllipsis = styled.div`
```
sophstad, line 16: [nit] Would a `<span>` be more appropriate here? Seems like we may not want to create block element if necessary (could always wrap the span in a div if needed!)

khelif96 (approved):
Great job!

PR #316, opened 2023-06-16
EVG-20206: Upgrade Storybook to v7.0.22

minnakt:
EVG-20206

### Description 
This PR upgrades Storybook.

Due to [Storybook #22511](https://www.github.com/storybookjs/storybook/issues/22511), I removed `"type": "commonjs"` from the package.json. I remember adding this at the time of creating the Parsley repo but I didn't have a special reason for doing so. We don't have this setting in Spruce's package.json, and I think it should be fine to remove.


khelif96 (approved):

minnakt:
confirmed that index.html on production task works properly

PR #331, opened 2023-07-05
EVG-19643 Use the currently deployed commit hash as the basis for the deploy script logs

khelif96:
EVG-19643

### Description 
Previously the deploy script would only use the diff in commits between the current git tag and the previous tag without considering what is actually deployed. So in the event of a revert, we would not return the correct commits since we would only look at the tags and not what is actually deployed in prod. 

This PR adds a step to download the commit hash from production prior to deploying. We use that commit hash as the basis for the email. If that file does not exist for whatever reason we fall back to the git tag method.

Example email:
image


scripts/deploy.sh
```diff
@@ -1,5 +1,18 @@
...
+# If the curl succeeds, then use the githash from the curl
+PREVIOUS_DEPLOYED_COMMIT=$(curl -s $BASE_URL/commit.txt)
+
+# Save the current commit hash to a file
+echo "Saving current commit hash to file"
+echo $PREVIOUS_DEPLOYED_COMMIT > bin/previous_deploy.txt
+
+# Create a file in dist/ that contains the current commit hash
+CURRENT_COMMIT_HASH=$(git rev-parse HEAD)
+echo $CURRENT_COMMIT_HASH > dist/commit.txt
```
sophstad, line 15: Writing a new file feels like a roundabout way of saving this information. Did you consider [a `<meta>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta)? Something like this in `index.html` could work:

```html
<meta name="version" content={REACT_APP_VERSION} />
```

There may be a more significant lift with fetching and parsing this, though. Interested in your thoughts!

khelif96: [We actually already do this](https://github.com/evergreen-ci/parsley/blob/05feeb479651bdab46342814aafd615f5ebd48ab/index.html#L7)!
 I considered using the meta tag but the lift seemed a bit too complex for a shell script. I would have to curl the HTML file and then figure out a way to parse that tag. The _simplest_ option is [to use a regex to parse the HTML](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags) but I don't think that would be as good of an idea as just reading a plain text file.

Do you think its something that would be worth trying?

sophstad: Yeah the regex does feel messier overall. This approach works for me!

scripts/email.sh
sophstad, line 1: [ShellCheck](https://www.shellcheck.net) returns a few style issues in this file

sophstad (changes requested):

scripts/email.sh
khelif96: Wow, great catch I had no idea this existed! Fixed most of the errors except the one about the $PATH not being printed to avoid logging a user's $PATH.

scripts/deploy.sh
sophstad, line 1: Oops, shellcheck here would be great too!

scripts/deploy.sh
```diff
@@ -1,9 +1,17 @@
+#!/bin/sh
```
sophstad, line 1: For consistency, should use a bash shebang to match `scripts/email.sh`
```suggestion
#!/bin/bash
```

scripts/deploy-production.js
```diff
@@ -90,6 +94,17 @@ const getCommitMessages = () =>
     );
   });
 
+const getCurrentlyDeployedCommit = () =>
+  new Promise((resolve, reject) => {
+    exec("sh scripts/get-current-deployed-commit.sh", (err, stdout) => {
```
sophstad, line 99: Should probably invoke this with bash (or use `sh` everywhere)
```suggestion
    exec("bash scripts/get-current-deployed-commit.sh", (err, stdout) => {
```

sophstad (approved):
Lgtm once these changes are made!

scripts/get-current-deployed-commit.sh
```diff
@@ -4,7 +4,18 @@ BASE_URL=$REACT_APP_PARSLEY_URL
 # First download the currently deployed commit hash from s3
 # If the curl fails, then use the previous tag from git
 # If the curl succeeds, then use the githash from the curl
+
+
+# Bash check for error 
+
```
sophstad, line 10: Should this be included?

khelif96: it should not thank you!

PR #336, opened 2023-07-12
EVG-19861: Fix filter crash bug

sophstad:
EVG-19861

### Description 

- A crash was occurring when a user introduced a change (via filter, etc.) that caused the row count to be less than the max page size while not on the zeroth page. This led to the page size being calculated as a negative number.
- When this occurs, return the row count as the total page size and reset the page index to 0.

### Testing 

- Added unit test that currently fails on `main`


khelif96 (approved):
Great job works great for me!

PR #337, opened 2023-07-12
EVG-20004 Trim log lines if they are too large

khelif96:
EVG-20004

### Description 
The origin of this bug was that a team was logging a single line that was over 180 MB which would cause performance problems when rendering. This PR limits the maximum size of a single log line to 5 MB which is 1MB more than what logkeepers limit is. 

image


sophstad (commented):
Do you think there would be any value to notifying a user if any log lines were trimmed? Maybe there's a chance teams would want to update their tooling to prevent them if they were informed.

khelif96:
> Do you think there would be any value to notifying a user if log lines were trimmed? Maybe there's a chance teams would want to update their tooling to prevent them if they were informed.

Good call I added a warning see the image.

src/constants/errors.ts
```diff
@@ -0,0 +1,6 @@
+import { getBytesAsString } from "utils/string";
+import { LOG_LINE_SIZE_LIMIT } from "./logs";
+
+export const LOG_LINE_TOO_LARGE_ERROR = `Parsley was unable to process one or more lines due to performance reasons since they exceed the line size limit of ${getBytesAsString(
```
sophstad, line 4: [nit] May be useful to reflect the severity of the message
```suggestion
export const LOG_LINE_TOO_LARGE_WARNING = `Parsley was unable to process one or more lines due to performance reasons since they exceed the line size limit of ${getBytesAsString(
```

src/utils/streams/index.ts
```diff
@@ -31,6 +33,16 @@ const decodeStream = async (stream: ReadableStream) => {
     }
 
     result.push(...lines);
+
+    if (lineSizeLimit) {
+      if (result[result.length - 1].length > lineSizeLimit) {
```
sophstad, line 38: For testing purposes I set the `lineSizeLimit` to 10 bytes and this line was really flaky, so the `trimmedLines` flag was only sometimes set. This seems like it is checking if the 2nd to last line of the log is too long? Also, what is the difference between the trim operation performed here and calling `trimLineToMaxSize` in `src/context/LogContext/state.ts`?

src/utils/string/index.ts
```diff
@@ -93,3 +95,16 @@ export const getBytesAsString = (bytes: number, decimals = 2) => {
...
+
+/**
+ * `trimLineToMaxSize` trims a line to the max size limit
+ * @param line - the line to trim
+ * @returns the trimmed line
+ * @example
+ */
+export const trimLineToMaxSize = (line: string) => {
+  if (line.length > LOG_LINE_SIZE_LIMIT) {
+    return `${line.substring(0, LOG_LINE_SIZE_LIMIT)}...`;
```
sophstad, line 107: [nit] Could use a single ellipsis character here instead of 3 periods
```suggestion
    return `${line.substring(0, LOG_LINE_SIZE_LIMIT)}…`;
```

src/utils/streams/index.ts
khelif96: Looks like the commit history on this pr is pretty messed up. I'm gonna close it and opened https://github.com/evergreen-ci/parsley/pull/342 instead.

khelif96:
Superseded by https://github.com/evergreen-ci/parsley/pull/342

PR #338, opened 2023-07-13
EVG-20172: Fix position of share icon

minnakt:
EVG-20172

### Description 
This PR makes the share icon position fixed (next to the line number). Also moved the icon to the right a bit.

### Screenshots

Screenshot 2023-07-12 at 9 02 11 PM

### DIFF
--- a/src/components/LogRow/BaseRow/index.tsx
+++ b/src/components/LogRow/BaseRow/index.tsx
@@ -104,7 +104,7 @@ const BaseRow: React.FC<BaseRowProps> = ({
       onDoubleClick={handleDoubleClick}
       shared={shared}
     >
-      <StyledIcon
+      <ShareIcon
         data-cy={`log-link-${lineNumber}`}
         glyph={shared ? "ArrowWithCircle" : "Link"}
         onClick={handleClick}
@@ -203,11 +203,12 @@ const RowContainer = styled.div<{
   }
 `;
 
-const StyledIcon = styled(Icon)`
+const ShareIcon = styled(Icon)`
   cursor: pointer;
   user-select: none;
   flex-shrink: 0;
-  align-self: center;
+  margin-left: ${size.xxs};
+  margin-top: 2px;
 `;
 
 const Index = styled.pre<{ lineNumber: number }>`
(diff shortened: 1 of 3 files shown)


minnakt:
I'm also kind of tempted to move the line numbers closer to the icon since I think it looks nicer...

Screenshot 2023-07-13 at 9 59 52 AM

src/components/LogRow/BaseRow/index.tsx
```diff
@@ -104,7 +104,7 @@ const BaseRow: React.FC<BaseRowProps> = ({
       onDoubleClick={handleDoubleClick}
       shared={shared}
     >
-      <StyledIcon
```
sophstad, line 107: Tbh I never knew what the function of this was button until now 😂 what do you think of adding a `title` attribute explaining it?

src/components/LogRow/BaseRow/index.tsx
sophstad, line 1: This is an existing problem but the link icon in the sidebar overflows awkwardly with super long line numbers (see [here](https://parsley.mongodb.com/resmoke/3813070f31d9eb165e2c85e0b221948c/test/175dda75e01390044d5f30c6b30a3c50?bookmarks=0,595200,1423186&filters=101%255C%257B.%2A%255C%257D&shareLine=1422606)). 
image

Fine with opening a new ticket or fixing here!

minnakt: I created [EVG-20442](https://jira.mongodb.org/browse/EVG-20442)!

src/components/LogRow/BaseRow/index.tsx
minnakt: I gave it a try but I don't think LG supports the `title` attribute properly, because when I hover no tooltip text ever appears =( I think the title text needs to be embedded in the original SVG, so it's hard to work around

minnakt: Tracked by [LG-3416](https://jira.mongodb.org/browse/LG-3416)

sophstad (approved):

PR #341, opened 2023-07-20
EVG-18913 Add lint rule to sort destructured keys

khelif96:
EVG-18913

### Description 
Adds key sorting for destructured objects


sophstad (commented):
Did you consider the [`sort-keys` ESLint rule](https://eslint.org/docs/latest/rules/sort-keys)? It's stricter than this one but if we're enforcing a sort I feel like it may be nice to be consistent. Plus it wouldn't require another external plugin.

khelif96:
> Did you consider the [`sort-keys` ESLint rule](https://eslint.org/docs/latest/rules/sort-keys)? It's stricter than this one but if we're enforcing a sort I feel like it may be nice to be consistent. Plus it wouldn't require another external plugin.

The official plugin does not support auto fixing so I found a fork of it with the same API that does.

src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -17,14 +17,14 @@ const ParsleyRow: RowRendererFunction = ({ logType, processedLogLines }) => {
...
-    scrollToLine,
     preferences,
     range,
+    scrollToLine,
     searchLine,
     searchState,
   } = useLogContext();
   const { searchTerm } = searchState;
-  const { wrap, prettyPrint } = preferences;
+  const { prettyPrint, wrap } = preferences;
```
sophstad, line 27: I was able to change the order of a destructured variable and commit without any warning/error appearing. I think we should be surfacing this rule at least during the lint-staged process. Are you able to reproduce?

khelif96: I don't think this rule actually affects destructured objects. I'll re-add the other rule to make this work.

sophstad (approved):
Works for me as long as you don't feel like these rules are too aggressive! We can always circle back if they are.

PR #342, opened 2023-07-20
EVG-20004 Trim log lines if they are too large

khelif96:
EVG-20004

### Description 
Reopening #337 because of some messed-up commits
The origin of this bug was that a team was logging a single line that was over 180 MB which would cause performance problems when rendering. This PR limits the maximum size of a single log line to 5 MB which is 1MB more than what logkeepers limit is. 

### Screenshots
image


src/utils/streams/index.ts
```diff
@@ -31,6 +35,16 @@ const decodeStream = async (stream: ReadableStream) => {
     }
 
     result.push(...lines);
+
+    if (lineSizeLimit) {
+      if (result[result.length - 1].length > lineSizeLimit) {
```
sophstad, line 40: Same comment here about this flaky condition. It doesn't actually iterate over every line (at least in [this log](https://parsley.mongodb.com/evergreen/spruce_ubuntu1604_test_2c9056df66d42fb1908d52eed096750a91f1f089_22_03_02_16_45_12/0/task?bookmarks=0,297) that I'm looking at). You can also see that the lines in the screenshot have not been trimmed to 10 bytes.

image

sophstad (changes requested):

src/utils/streams/index.ts
khelif96: oof sorry missed this when recreating the pr.

khelif96: Thanks for catching this. I just updated the pr

src/hooks/useLogDownloader/index.ts
```diff
@@ -71,12 +72,31 @@ const useLogDownloader = (
...
+          });
+          reportError({
+            message: "Log line size limit exceeded",
+            metadata: {
+              trimmedLines,
+              url,
+            },
+            name: "Log download incomplete",
+          }).warning();
+          dispatchToast.warning(LOG_LINE_TOO_LARGE_WARNING, true, {
```
sophstad, line 96: Two toasts are firing every time I load the page
image

src/hooks/useLogDownloader/index.ts
```diff
@@ -71,12 +72,31 @@ const useLogDownloader = (
...
+            title: "Log not fully downloaded",
```
sophstad, line 97: What do you think about adding `shouldTimeout: false`? I feel like this isn't a transient notice like e.g. a test starting that will be easily observed elsewhere on the page, so we should make it possible for users to digest the toast before it's gone forever.

src/constants/errors.ts
```diff
@@ -0,0 +1,6 @@
+import { getBytesAsString } from "utils/string";
+import { LOG_LINE_SIZE_LIMIT } from "./logs";
+
+export const LOG_LINE_TOO_LARGE_WARNING = `Parsley was unable to process one or more lines due to performance reasons since they exceed the line size limit of ${getBytesAsString(
```
sophstad, line 4: Performance reasons feels implied, may as well keep the toast concise
```suggestion
export const LOG_LINE_TOO_LARGE_WARNING = `Parsley was unable to process one or more lines since they exceed the line size limit of ${getBytesAsString(
```

sophstad (changes requested):

src/hooks/useLogDownloader/index.ts
khelif96: This is due to React strict mode when running locally. The network request actually fires twice under the hood and thus triggers the toast callback twice. It wouldn't happen on a production build.

khelif96: image

src/hooks/useLogDownloader/index.ts
khelif96: That makes sense done!

src/utils/string/index.ts
```diff
@@ -93,3 +93,17 @@ export const getBytesAsString = (bytes: number, decimals = 2) => {
   const i = Math.floor(Math.log(bytes) / Math.log(k));
   return `${parseFloat((bytes / k ** i).toFixed(dm))} ${sizes[i]}`;
 };
+
+/**
+ * `trimLogLineToMaxSize` trims a line to the max size limit
+ * @param line - the line to trim
+ * @param maxSize - the max line size limit
+ * @returns the trimmed line
+ * @example
```
sophstad, line 102: [nit] Remove (or add an example!)

sophstad (approved):

PR #343, opened 2023-07-21
EVG-20443: Resolve Evergreen test log URLs via GQL when possible

SupaJoon:
EVG-20443

### Description 
These code changes partly implement the decision to [Use GraphQL to Resolve Test Log URLs](https://github.com/evergreen-ci/parsley/blob/main/docs/06-decisions/2023-07-18_resolving_test_log_url.md) in a backwards compatible way. The decision will be fully implemented when resmoke logs are supported in EVG-20023. The TestResult GQL query from this PR will show results after the code changes for this EVG-20455 is deployed. The fallback URL generation is kept for backwards compatibility.


src/pages/LogView/LoadingPage/useResolveLogURL.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+    case LogTypes.EVERGREEN_TEST_LOGS: {
+      if (!taskID || !execution || !testID || isLoadingTest) {
+        break;
+      }
+      const { url, urlLobster, urlRaw } =
+        testData?.task?.tests.testResults[0]?.logs || {};
+      rawLogURL =
+        urlRaw ??
+        getEvergreenTestLogURL(taskID, execution, testID, {
```
khelif96, line 96: My question is would we ever expect this to be happen? If the test resolver doesn't return anything would we expect getEvergreenTestLogURL to point to a real log?

src/pages/LogView/LoadingPage/useResolveLogURL.ts
```diff
@@ -0,0 +1,121 @@
...
+  const { data: testData, loading: isLoadingTest } = useQuery<
+    TestLogUrlQuery,
+    TestLogUrlQueryVariables
+  >(GET_TEST_LOG_URL, {
+    variables: {
+      testName: `^${testID}$`,
+      taskID: taskID as string,
+      execution: parseInt(execution as string, 10),
+    },
+    skip: !(
```
khelif96, line 46: One of the biggest motivations of https://jira.mongodb.org/browse/EVG-19354 is that all test logs would originate from the same source and we would resolve the logs the same way from the test results resolver. So this should apply to both resmoke and normal test logs.

khelif96 (changes requested):

src/pages/LogView/LoadingPage/useResolveLogURL.ts
khelif96: Sorry about the blocking review this was more of a comment-style review but I clicked the wrong thing.

khelif96 (approved):
LGTM

PR #344, opened 2023-07-24
EVG-20503 Verify bin directory exists before writing to file

khelif96:
EVG-20503

### Description 
The spruce and parsley deploy scripts have a step where they write to a file called bin/previous_deploy.txt.  While this file usually always exists if a person had developed in these repositories it would not if it is a fresh repo. This adds a check to verify it exists


scripts/get-current-deployed-commit.sh
```diff
@@ -25,6 +25,12 @@ then
...
 
+# Check if bin directory exists
+if [ ! -d "bin" ]
+then
+    mkdir bin
+fi
+
 # Save the current commit hash to a file
 echo "$PREVIOUS_DEPLOYED_COMMIT"
 echo "$PREVIOUS_DEPLOYED_COMMIT" > bin/previous_deploy.txt
```
minnakt, line 36: I notice you often run into this problem, it may be a good idea to enable "insert final new line on save" in VSCode so that this won't happen

minnakt (approved):

PR #346, opened 2023-07-31
EVG-18978: Prevent highlights from corrupting HTML tags

minnakt:
EVG-18978

### Description 
This PR moves the highlight logic into the HTML parser to prevent highlights from corrupting HTML tags.

This approach solves a number of bugs in production including:
* Highlights can mess up log lines that include `<` and `>` since they are escaped to `&gt;` and `&lt;` respectively.
 
Correct line      |  Incorrect line after searching for `g`
:-------------------------:|:-------------------------:
gt_right | gt_wrong

* Searching for something like `span` will corrupt the `span` tags added via `ansi_to_html`.

Correct line      |  Incorrect line after searching for `span`
:-------------------------:|:-------------------------:
span_right | span_wrong

* Inconsistencies between searching and highlighting because of escaped vs unescaped text.

Search works, but missing highlight    |  Highlight works, but the search term needs to be escaped
:-------------------------:|:-------------------------:
newbranch-highlight | newbranch-search

### Testing 
- Added e2e & unit tests

### DIFF
--- a/src/components/LogRow/BaseRow/Highlighter.tsx
+++ b/src/components/LogRow/BaseRow/Highlighter.tsx
@@ -1,5 +1,5 @@
 import { memo, useMemo } from "react";
-import renderHtml from "utils/renderHtml";
+import { highlightHtml } from "utils/renderHtml";
 
 interface HighlighterProps {
   ["data-cy"]?: string;
@@ -19,7 +19,7 @@ const Highlighter: React.FC<HighlighterProps> = memo((props) => {
   } = props;
 
   const htmlToRender = useMemo(
-    () => renderHtml(text, searchTerm, highlights),
+    () => highlightHtml(text, searchTerm, highlights),
     [text, searchTerm, highlights]
   );
 
--- a/src/components/LogRow/BaseRow/Highlighter.tsx
+++ b/src/components/LogRow/BaseRow/Highlighter.tsx
@@ -1,5 +1,5 @@
 import { memo, useMemo } from "react";
-import { highlightHtml } from "utils/renderHtml";
+import highlightHtml from "utils/highlightHtml";
 
 interface HighlighterProps {
   ["data-cy"]?: string;
--- a/src/components/LogRow/BaseRow/Highlighter.tsx
+++ b/src/components/LogRow/BaseRow/Highlighter.tsx
@@ -1,0 +1,34 @@
+import { memo, useMemo } from "react";
+import renderHtml from "utils/renderHtml";
+
+interface HighlighterProps {
+  ["data-cy"]?: string;
+  children: string;
+  color?: string;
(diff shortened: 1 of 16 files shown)


src/components/LogRow/BaseRow/Highlighter.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface HighlighterProps {
+  ["data-cy"]?: string;
+  children: string;
+  color?: string;
+  highlights?: RegExp;
+  searchTerm?: RegExp;
+}
+
+const Highlighter: React.FC<HighlighterProps> = memo((props) => {
```
khelif96, line 12: Can you write some unit tests for this as well?

src/utils/renderHtml/index.tsx
```diff
@@ -58,4 +61,59 @@ const renderHtml = (html: string = "", options: renderHtmlOptions = {}) => {
...
 };
 
+/**
+ * `highlightHtml` adds highlights in the form of <mark> tags.
+ * @param html - The html string to parse
+ * @param searchTerm - The active search term as a regex expression
+ * @param highlights - The active highlights as a regex expression
+ * @returns - html string converted to an array of domnodes with highlighted text
+ */
+export const highlightHtml = (
```
khelif96, line 71: I would argue that the logic here is complicated enough to deserve its own folder/file.

src/utils/renderHtml/renderHtml.test.tsx
```diff
@@ -51,6 +50,43 @@ describe("renderHtml", () => {
   });
 });
 
+describe("highlightHtml", () => {
```
khelif96, line 53: Can we add some tests for the color logic since that is now included here.

khelif96: Also tests for the overlap logic.

khelif96 (changes requested):

khelif96 (changes requested):
Sorry accidentally requested myself. Feel free to request when you are ready.

src/components/LogRow/BaseRow/Highlighter.tsx
minnakt: I moved some tests out of `Row.test.tsx` into these other files because it was getting really redundant (i.e. if `Highligher` is testing these things, and `Row` contains `Highlighter`, then `Row` probably doesn't need to test it again)

src/utils/renderHtml/index.tsx
minnakt: Added lots of file changes as I had to move some stuff around to make this work

khelif96 (approved):
Great job! I would recommend some integration tests to have some coverage of when these components are used together.

PR #361, opened 2023-08-14
EVG-20617 Only upload to cypress if on a pr, patch or mainline commit

khelif96:
EVG-20617

### Description 
Following up on our conversation this should reduce unnecessary uploads to the Cypress cloud service.


sophstad (approved):
Nice!

PR #362, opened 2023-08-17
EVG-20581 Don't auto-dismiss log download incomplete toast

khelif96:
EVG-20581


minnakt (approved):
nice! if it's not too much trouble, while you are fixing the tests, could you also fix the weird spacing issue going on [here](https://github.com/evergreen-ci/parsley/blob/main/src/context/toast/index.tsx#L53-L72)?

PR #363, opened 2023-08-17
EVG-19795: Load log page instead of infinite loading bar for empty logs.

SupaJoon:
EVG-19795

### Description 
The `hasLogs` conditional prevents rendering the Log UI for empty logs so I based the render off of whether logs were ingested. The screenshot is from a log file I seeded in my local evergreen.
The cypress test relies on this PR: https://github.com/evergreen-ci/evergreen/pull/6962
### Screenshots
Screenshot 2023-08-22 at 3 40 42 PM

### DIFF
--- a/src/gql/generated/types.ts
+++ b/src/gql/generated/types.ts
@@ -60,6 +60,16 @@ export type Annotation = {
   webhookConfigured: Scalars["Boolean"];
 };
 
+export enum Arch {
+  Linux_64Bit = "LINUX_64_BIT",
+  LinuxArm_64Bit = "LINUX_ARM_64_BIT",
+  LinuxPpc_64Bit = "LINUX_PPC_64_BIT",
+  LinuxZseries = "LINUX_ZSERIES",
+  Osx_64Bit = "OSX_64_BIT",
+  OsxArm_64Bit = "OSX_ARM_64_BIT",
+  Windows_64Bit = "WINDOWS_64_BIT",
+}
+
 export enum BannerTheme {
   Announcement = "ANNOUNCEMENT",
   Important = "IMPORTANT",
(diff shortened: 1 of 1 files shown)


src/pages/LogView.tsx
```diff
@@ -8,9 +8,9 @@ interface LogViewProps {
 }
 
 const LogView: React.FC<LogViewProps> = ({ logType }) => {
-  const { hasLogs } = useLogContext();
+  const { hasIngestedLogs } = useLogContext();
 
-  return hasLogs ? (
+  return hasIngestedLogs ? (
```
sophstad, line 13: It may be nice to add a different view when the logs are empty to indicate that it's not just an error rendering the logs. There's an [empty state LeafyGreen component](https://www.mongodb.design/component/empty-state/example/), or you could do something simpler

src/pages/LogView.tsx
```diff
@@ -8,9 +8,9 @@ interface LogViewProps {
...
 
 const LogView: React.FC<LogViewProps> = ({ logType }) => {
-  const { hasLogs } = useLogContext();
+  const { hasLogs, logsAreEmpty } = useLogContext();
 
-  return hasLogs ? (
+  return hasLogs || logsAreEmpty ? (
     <LogWindow isUploadedLog={false} logType={logType} />
   ) : (
     <LoadingPage logType={logType} />
```
minnakt, line 16: Hmm it's somewhat confusing to me how there is `hasLogs` and `logsAreEmpty`, since they sound like they represent the same thing. One thing I think might be more intuitive would be to have a `loading` boolean that starts as true and gets set to false in `INGEST_LOGS`. That way, you could flip this condition:

```
  const { loading } = useLogContext();

  return loading ? (
    <LoadingPage logType={logType} />
  ) : (
    <LogWindow isUploadedLog={false} logType={logType} />
  );
```

I removed `logsAreEmpty` and tried it this way and I think it worked with your existing changes! Pls feel free to DM me if this doesn't make sense haha 😅

sophstad (approved):

minnakt (approved):

PR #373, opened 2023-09-05
EVG-18748: Remove selectedLine redirect

sophstad:
EVG-18748

### Description 

Stop supporting `selectedLine` query param in Parsley URLs now that the TTL has been reached.

### Testing 

- Remove tests for redirect support


khelif96 (approved):
🚀

PR #374, opened 2023-09-06
EVG-20175: Avoid calling controller.close() twice

minnakt:
EVG-20175

### Description 
This part of the code ends up calling `controller.close()` two times, because when it breaks out of the loop it also calls `controller.close()`. I just removed the extra call.

Also added a breadcrumb. (Which will get Sentry-fied in [EVG-20825](https://jira.mongodb.org/browse/EVG-20825))

### Testing 
- Manual testing. Set the file limit to something low (like 2) and uploaded a file and made sure the error went away.


khelif96 (approved):
Great catch!

PR #395, opened 2023-09-28
EVG-20974: Downgrade graphql dependency

sophstad:
EVG-20974

### Description 

- `yarn dev` is broken locally due to [this syntax error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Missing_name_after_dot_operator?utm_source=mozilla&utm_medium=firefox-console-errors&utm_campaign=default) included in the graphql package. Reverting the upgrade in https://github.com/evergreen-ci/parsley/pull/386 fixes it.
- I do not know why this package works fine in Spruce 🤔 but would rather figure that out separately

### Testing 

- `yarn dev` runs locally

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@leafygreen-ui/tooltip": "10.0.10",
     "@leafygreen-ui/typography": "16.5.0",
     "ansi_up": "5.1.0",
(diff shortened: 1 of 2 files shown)


minnakt (approved):
yeehaw!

PR #407, opened 2023-10-09
EVG-19872: Apply user's wrap setting after log pane has loaded

minnakt:
EVG-19872

### Description 
This PR makes it so that the user's wrap setting is applied after the log pane has initially loaded. 

It seems like the virtualization library can't measure the row heights correctly if wrap is enabled when loading a large log. You can try this on production with the log linked in the ticket. Scroll to index has a similar problem so I just added the wrap logic in the same place.

### Testing 
- Added Jest tests to make sure the functions are properly called
- Since this bug only occurs with large logs, it's easier to test this manually. The changes are deployed on beta at the moment so feel free to use that for testing.


src/components/LogPane/index.tsx
```diff
@@ -11,40 +13,42 @@ interface LogPaneProps {
...
+        } else {
+          leaveBreadcrumb(
+            "shareLine not provided or found in processedLogLines",
+            { shareLine },
+            "process"
+          );
+        }
+
+        // Wrap can be enabled after the log pane has initially loaded.
+        setWrap(Cookies.get(WRAP) === "true");
```
sophstad, line 47: Should we only call this if necessary?
```suggestion
        if (Cookies.get(WRAP) === "true") {
          setWrap(true);
        }
```

sophstad (approved):

PR #424, opened 2023-10-31
DEVPROD-1647 Add animation to details button when details menu items are changed.

khelif96:
DEVPROD-1647

### Description 
Adds an animation to the details menu item when ever the search range is changed.

### Screenshots

https://github.com/evergreen-ci/parsley/assets/4605522/2a4b8043-1a75-42ff-9903-930c03453f7a


sophstad (approved):
Looks very cool 😎

sophstad (changes requested):
Oops I take it back! The animation still applies when a user manually changes the range which seems a bit aggressive.

khelif96:
> Oops I take it back! The animation still applies when a user manually changes the range which seems a bit aggressive.

Good call!

sophstad (changes requested):
It seems like the animation also applies on initial load now

https://github.com/evergreen-ci/parsley/assets/9298431/562ab60c-7737-494b-88fa-0918de619cce

khelif96:
> It seems like the animation also applies on initial load now
> 
>  Screen.Recording.2023-11-07.at.10.23.42.AM.mov

I pushed up a fix that will prevent it from happening on production builds. Due to how react 18 renders things twice in dev mode the animation will only display on first render locally but shouldn't on a production build. (You can disable strict mode to test this)

khelif96:
> Single line search in range selection isn't updating the details menu nor showing the animation. Those code changes can be made in a separate PR if you like. 
That is intentional.
 https://github.com/evergreen-ci/parsley/blob/9a8a8d7e78bc2266b03804dcec00c8f209250140/src/components/LogRow/BaseRow/SharingMenu/index.tsx#L59

We can open a ticket to update it if the flow doesn't make sense. 

> I'm unable to figure out how to reset the search in range selection without opening the details menu. Is it possible?

That is not possible. When we originally opened this pr the UX didn't make sense to reset the search range from the selection menu. It is why this pr exists so that users are made aware of the change in the details menu.

sophstad (approved):

PR #430, opened 2023-11-06
DEVPROD-2232 Log download error at lower severity

khelif96:
DEVPROD-2232

### Description 
Since these alerts have been pretty noisy, let's log them as a warning and follow up by adjusting the alerting criteria in sentry in case they spike.


sophstad (approved):
🙌

PR #431, opened 2023-11-08
DEVPROD-929: Use task log links from API

sophstad:
DEVPROD-929

### Description 

- Use log URLs returned by the `task` resolver to generate download/HTML/raw links.
- Update 404 logic to handle invalid tasks

### Testing 

- Add unit test


src/pages/LogView/LoadingPage/useResolveLogURL.ts
```diff
@@ -189,7 +197,7 @@ export const useResolveLogURL = ({
     htmlLogURL,
     jobLogsURL,
     legacyJobLogsURL,
-    loading: isLoadingTest,
+    loading: isLoadingTest || taskLoading,
```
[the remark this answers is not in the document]
minnakt: random observation but I wonder why we don't also include `isLoadingTaskFileData` here 🤔

src/hooks/useLogDownloader/index.ts
```diff
@@ -142,14 +144,19 @@ const useLogDownloader = (
           });
           setIsLoading(false);
         });
+    } else if (!url && !isLoadingEvergreen) {
+      dispatchToast.error("Log URL not specified, unable to download.", true, {
+        shouldTimeout: false,
+      });
+      setIsLoading(false);
```
minnakt, line 151: Possibly a silly question but would you mind describing what this part does? When I tried removing it, I could still load the logs 😮

src/hooks/useLogDownloader/index.ts
```diff
@@ -33,7 +34,8 @@ import { getBytesAsString } from "utils/string";
 const useLogDownloader = (
   url: string,
   logType: LogTypes,
-  downloadSizeLimit: number = LOG_FILE_SIZE_LIMIT
+  downloadSizeLimit: number = LOG_FILE_SIZE_LIMIT,
+  isLoadingEvergreen: boolean = false
 ) => {
```
minnakt, line 39: May be worth making these arguments into an object, since it's a bit difficult to read when it's being called in other files

src/gql/queries/get-logkeeper-task.graphql
```diff
@@ -5,6 +5,12 @@ query LogkeeperTask($buildId: String!) {
     id
     task {
       ...BaseTask
+      logs {
+        agentLogLink
+        allLogLink
+        systemLogLink
+        taskLogLink
+      }
```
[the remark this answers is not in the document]
sophstad: For the sake of typing I found it was significantly cleaner to include these fields (specifically for the typing of `useTaskQuery()`.

If you don't object, I'll add them to `BaseTask`

src/pages/LogView/LoadingPage/useResolveLogURL.ts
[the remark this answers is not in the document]
sophstad: Good point, added it!

src/hooks/useLogDownloader/index.ts
sophstad: You can see the failed run [here](https://spruce.mongodb.com/task/parsley_ubuntu2204_large_e2e_test_patch_d9645fc8c24e8bed33c4309c376eb80360adc3b0_654bbfaa61837db75b414fa0_23_11_08_17_04_43?execution=0&sortBy=STATUS&sortDir=ASC); basically if there's no URL to download, we never enter the `if (url)` block, and Parsley is stuck in an infinite "downloading" state.

It would probably make more sense to only show that state on Parsley once the download actually begins/a URL is present, but as-is this was a more straightforward fix.

src/hooks/useLogDownloader/index.ts
```diff
@@ -142,14 +153,19 @@ const useLogDownloader = (
           });
           setIsLoading(false);
         });
+    } else if (!url && !isLoadingEvergreen) {
+      dispatchToast.error("Log URL not specified, unable to download.", true, {
+        shouldTimeout: false,
+      });
```
minnakt, line 159: Ah ok, does it make sense to use this existing pattern in the file?

```suggestion
      setError("Log URL not specified, unable to download.");
```

minnakt (approved):
nice 😃

PR #435, opened 2023-11-14
DEVPROD-2643: Remove useLogDownloader hook dependency

sophstad:
DEVPROD-2643

### Description 

- Remove `useLogDownloader` dependency on `isEvergreenLoading`. This dependency caused the hook to run multiple times; when the bug was deployed to production you could observe multiple requests to Logkeeper. I think the subsequent request to Logkeeper also resulted in the first one being terminated, hence the "Incomplete log" warnings we saw earlier today.

### Testing 

- Verified on beta that removing this dependency results in a single call to fetch the log. It's deployed to beta right now if you care to look!


sophstad:
Reintroducing reverted code in #436

PR #436, opened 2023-11-15
DEVPROD-2643: Fix useLogDownloader bug

sophstad:
DEVPROD-2643

### Description
- First commit reverts #434 i. e. reintroduces #431 
- Second commit fixes bug
  - Remove useLogDownloader dependency on isEvergreenLoading. This dependency caused the hook to run multiple times; when the bug was deployed to production you could observe multiple requests to Logkeeper. I think the subsequent request to Logkeeper also resulted in the first one being terminated, hence the "Incomplete log" warnings we saw earlier today.
  - In the case of an invalid task URL (i.e. `/evergreen/invalid/0/task`) where the GraphQL query doesn't return a download link, use the previous approach to construct a link given the task ID, execution, etc.

### Testing
- Verified on beta that removing this dependency results in a single call to fetch the log. It's deployed to beta right now if you care to look!


src/constants/logURLTemplates.ts
```diff
@@ -101,40 +103,26 @@ const getResmokeLogURL = (
...
- * @returns an Evergreen URL of the format `/task/${taskID}/${execution}?type=${OriginToType[origin]}&text=true`
- */
 const getEvergreenTaskLogURL = (
-  taskID: string,
-  execution: string | number,
-  origin: keyof typeof OriginToType,
-  options: { priority?: boolean; text?: boolean }
+  logLinks: TaskType["logs"],
+  origin: string,
+  params: { [key: string]: any } = {}
```
minnakt, line 116: does it make sense to type the params a bit more strongly?

src/hooks/useLogDownloader/index.ts
```diff
@@ -17,28 +17,35 @@ import {
 import { fetchLogFile } from "utils/fetchLogFile";
```
minnakt, line 17: it's not a big deal, but do you know where this flash of 0 comes from? (only happens on bigger logs)

https://github.com/evergreen-ci/parsley/assets/47064971/5a98864c-3cdb-4eb2-babd-c492171cbdb3

src/hooks/useLogDownloader/index.ts
```diff
@@ -142,6 +150,8 @@ const useLogDownloader = (
           });
           setIsLoading(false);
         });
+    } else if (url === null) {
+      setError("Log URL not specified, unable to download.");
```
minnakt, line 154: I think i might tweak the wording here to something along the lines of this:
```suggestion
      setError("Could not retrieve log URL from task, unable to download.");
```

minnakt (commented):
sorry for the late review 😭 pls ping me directly whenever you want a re-review!

src/constants/logURLTemplates.ts
sophstad: This seemed to be the general typing we use for query params, à la `useQueryParams` https://github.com/evergreen-ci/parsley/blob/4126a3928d37da614d211fb0ecb3ef2735ac0ab6/src/hooks/useQueryParam/index.ts#L17

src/hooks/useLogDownloader/index.ts
sophstad: I actually think this is a React strict-mode quirk that occurs locally and results in certain things firing twice! We've been running into more of those lately. I can reproduce this on `main` when running locally, and don't see the behavior when my branch is deployed to beta.

src/pages/LogView/LoadingPage/useResolveLogURL.test.tsx
```diff
@@ -129,6 +175,7 @@ describe("useResolveLogURL", () => {
       }
     );
     await waitForNextUpdate();
+    await waitForNextUpdate();
```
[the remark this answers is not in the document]
sophstad: My mistake thank you!

src/pages/LogView/LoadingPage/useResolveLogURL.ts
```diff
@@ -142,19 +151,32 @@ export const useResolveLogURL = ({
...
+            text: true,
+          })
+        : constructEvergreenTaskLogURL(taskID, execution, origin as any, {
+            text: true,
+          });
+      htmlLogURL = task?.logs
+        ? getEvergreenTaskLogURL(task.logs, origin, {
+            text: false,
+          })
+        : constructEvergreenTaskLogURL(taskID, execution, origin as any, {
```
minnakt, line 177: wondering if it makes sense to copy the types for `getEvergreenTaskLogURL` to remove the `origin as any` stuff!

src/pages/LogView/LoadingPage/useResolveLogURL.test.tsx
```diff
@@ -184,7 +254,7 @@ const getEmptyTestLogURLMock: ApolloMock<
     variables: {
       execution: 0,
       taskID: "a-task-id",
-      testName: "^a-test-name$",
+      testName: "^a-test-name-that-doesnt-exist$",
```
minnakt, line 257: ty for fixing 😁

minnakt (approved):
lgtm!

PR #442, opened 2023-11-20
DEVPROD-818: Remove @testing-library/react-hooks

minnakt:
DEVPROD-818

### Description 
Removes `@testing-library/react-hooks` dependency. `renderHook` is now accessible via `@testing-library/react`.

### Testing 
- Jest tests should pass


minnakt:
merging because failing test is flaky

PR #447, opened 2023-11-29
DEVPROD-1629: Preserve server prefix when processing resmoke logs

minnakt:
DEVPROD-1629

### Description 
Resmoke logs can have multiple pid/ports in the prefix.

Timing on large log (1927870 lines):
* Preprocessing time on main: 2962.50085449 ms
* Preprocessing time on this PR: 3310.687744140625 ms

An extra ~300 ms doesn't seem too bad for a large log.

### Note 
It's possible to achieve better performance by using the regex `/ ([sdbc]+\d{1,5})\|/` rather than `/ ([shdbc]+\d{1,5})\|/g` when building the color map in `getColorMapping`. This regex ignores the `sh` prefix and only reads the second prefix. However, this makes assumptions about resmoke logs that could be unsafe to make (e.g. that there always be maximum two prefixes, that the second prefix will never start with `sh`). It also means that lines with only the `sh` prefix will lack a color.

### Screenshots
Screenshot 2023-11-29 at 5 59 37 PM

### Testing 
* On test log; see JIRA ticket
* Jest tests


src/utils/resmoke/helpers.ts
```diff
@@ -3,16 +3,22 @@
...
- * `getPort` returns the port associated with a resmoke line
+ * `getPorts` returns the ports associated with a resmoke line
  * @param line - the resmoke line
- * @returns the port associated with a resmoke line
+ * @returns Ports associated with a resmoke line. There can be 1 or more ports.
  */
-const getPort = (line: string) => {
-  const port = line.match(portRegex)?.[1];
-  return port;
+const getPorts = (line: string) => {
```
sophstad, line 13: Just so it's crystal clear we're not returning an array of ports
```suggestion
const getPorts = (line: string): string => {
```

sophstad (approved):

PR #456, opened 2023-12-13
DEVPROD-815 Upgrade node version to 20.10.0

khelif96:
DEVPROD-815

### Description
Bumps node to latest LTS version


sophstad (approved):
LGTM 👍 should we get rid of the `node_path` variable in Parsley's repo settings?

PR #457, opened 2023-12-15
DEVPROD-2680: Update New Relic snippet and remove Sentry Replay

minnakt:
[DEVPROD-2680](https://jira.mongodb.org/browse/DEVPROD-2680)

### Description 
This PR does two things: 
* Updates the agent string, as per advice of New Relic support.
* Reverts the integration of Sentry replays.

The removal of Sentry replays is a more of an educated guess than a definitive solution. I found some GitHub issues in the `sentry-javascript` repo where users reported performance issues for replays on large tables/grids, which seemed relevant. I also asked Mohamed to test on his machine and the memory usage was looking better.

Even if removing Sentry replays doesn't help the memory leak, it does solve a separate problem of not being able to load large logs (see comment in JIRA ticket). Removing Sentry replays allows this load to log properly.

### Testing 
If you'd like to test on beta, you'll need to change the conditional in `index.html` to include the beta environment. You can type `newrelic` in the console to confirm that New Relic is active (it should return an object rather than `undefined`). But be aware that you may not be able to replicate the memory leak in the first place, since it is highly inconsistent among users.


sophstad (approved):
LGTM. I would also be open to removing Sentry Replay from Spruce as MongoDB does not have enough replays included in our plan for it to be useful and so would prefer to minimize performance risks.

PR #463, opened 2024-01-10
DEVPROD-809: Upgrade Prettier to v3.1.1

minnakt:
DEVPROD-809

### Description 
Upgrades Prettier and re-ran `yarn eslint:fix` to fix any errors.

The reason you now see a bunch of trailing commas can be found from the [release notes](https://prettier.io/blog/2023/07/05/3.0.0.html#change-the-default-value-for-trailingcomma-to-all-11479httpsgithubcomprettierprettierpull11479-by-fiskerhttpsgithubcomfisker-13143httpsgithubcomprettierprettierpull13143-by-sosukesuzukihttpsgithubcomsosukesuzuki):

Screenshot 2024-01-10 at 12 13 29 PM

We currently use all of the defaults from Prettier so I think it makes sense to use this default as well.

### Testing 
- Made sure my changes still get linted properly on save in VSCode


sophstad (approved):
comma city over here

PR #470, opened 2024-01-17
DEVPROD-825 Remove references to Lobster

khelif96:
DEVPROD-825
DO NOT MERGE until 1/26
### Description 
Removes all outward facing links to lobster from parsley


src/components/DetailsMenu/DetailsMenuCard/ButtonRow/index.tsx
```diff
@@ -114,17 +114,6 @@ const ButtonRow: React.FC = () => {
...
-          data-cy="lobster-button"
-          disabled={!lobsterURL}
-          glyph={<Icon glyph="Export" />}
-          href={lobsterURL || ""}
-          onClick={() => sendEvent({ name: "Opened Lobster Logs" })}
-          target="_blank"
-        >
-          Lobster
-        </MenuItem>
         {legacyJobLogsURL && (
```
minnakt, line 117: If there's no job logs, the menu now can be empty and look weird for non-resmoke logs:
Screenshot 2024-01-26 at 10 17 04 AM

I think the easiest way to change this would be to move `legacyJobLogsURL` and wrap it as a conditional around the whole` <Menu />` or disable the `<MenuItem />`

PR #482, opened 2024-02-08
DEVPROD-3178: Support svc key in resmoke logs

minnakt:
DEVPROD-3178

### Description 
This PR adds support for the svc key.

### Screenshots
User's suggestion:
> I would just want us to add something like [S] for shard, [R] for router, or - for none/unknown, somewhere between the component and the context.

Screenshot 2024-02-08 at 1 01 47 PM

### Testing 
- Jest tests
- Tested against production; see ticket for example log. Note that not all resmoke logs use the svc key.


src/utils/resmoke/helpers.ts
```diff
@@ -156,6 +156,25 @@ const getPid = (line: string) => {
...
+/**
+ * `getSvc` returns the svc associated with a mongod process. It can be router-role work (R),
+ * shardserver-role work (S), or neither (-). It is not always present in resmoke logs.
+ * @param line - the resmoke line
+ * @returns the svc associated with the resmoke line
+ */
+const getSvc = (line: string) => {
+  const svc = line.match(svcRegex)?.[1];
+  return svc;
+};
```
SupaJoon, line 176:
```suggestion
const getSvc = (line: string) => line.match(svcRegex)?.[1];
```

src/utils/resmoke/index.ts
```diff
@@ -50,15 +51,16 @@ const processResmokeLine = (line: string) => {
...
   const pid = getPid(json)?.padEnd(7, " ") ?? "";
   const ctx = getContext(json) ?? "";
   const msg = getMessage(json) ?? "";
   const attr = getAttributes(json) ?? "";
 
-  const output = `${resmokeFunction} ${ports}| ${timeStamp} ${shellPrefix} ${configSrv} ${pid} [${ctx}] "${msg}"${
-    attr ? `,${attr}` : ""
-  }`;
+  const output = `${resmokeFunction} ${ports}| ${timeStamp} ${shellPrefix} ${configSrv}${
+    svc ? ` [${svc}]` : ""
```
SupaJoon, line 62: Do you think the alignment metadata alignment can throw people off when some lines contain svc and some don't? This leads me to wonder, if there is a benefit to making lifted variables configurable?

SupaJoon (approved):

src/utils/resmoke/index.ts
minnakt: I think making the variables configurable would be a cool project idea! For this key, I just made it so that the non-svc logs look as they do now, in hopes of reducing any confusion

minnakt: Oh wait I think I misinterpreted your question! I'll ask about this in the UI channel on Monday I think

PR #486, opened 2024-02-12
DevProd-1435: Use renderType GQL field to determine Row rendering format

SupaJoon:
DevProd-1435

### Description 
The resmoke route is not fully deprecated yet so we must support it until Julian lets us know. `renderingType` is only applicable to test logs at the moment.


src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -49,7 +50,15 @@ const ParsleyRow: RowRendererFunction = ({ logType, processedLogLines }) => {
         />
       );
     }
-    const Row = rowRendererMap[logType];
+
+    // At this point, logMetadata.renderingType is guaranteed to be defined from LogView/LoadingPage.tsx.
+    // logType can be removed from this calculation when the Parsley resmoke route is fully deprecated.
+    const Row =
+      logMetadata?.renderingType === "resmoke" ||
```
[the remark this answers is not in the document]
khelif96: I also don't quite understand this bit of logic. Is there a way we can preserve rowRenderMap it seems far more flexible and would allow the expansion of future row types in the future.

src/pages/LogDrop/FileDropper.tsx
```diff
@@ -42,7 +42,12 @@ const FileDropper: React.FC = () => {
   const onParse = useCallback(
     (logType: LogTypes | undefined) => {
       if (logType) {
-        setLogMetadata({ isUploadedLog: true, logType });
+        setLogMetadata({
+          isUploadedLog: true,
+          logType,
+          renderingType:
+            logType === LogTypes.RESMOKE_LOGS ? "resmoke" : "default",
```
khelif96, line 49: Same comment about using an enum value.

src/pages/LogDrop/ParseLogSelect/index.tsx
```diff
@@ -14,13 +14,18 @@ interface ParseLogSelectProps {
...
+
 const ParseLogSelect: React.FC<ParseLogSelectProps> = ({
   fileName,
   onCancel,
   onParse,
 }) => {
-  const [logType, setLogType] = useState<LogTypes | undefined>(
-    (Cookie.get(LAST_SELECTED_LOG_TYPE) as LogTypes) ?? undefined,
+  const [logType, setLogType] = useState<SelectState>(
+    (Cookie.get(LAST_SELECTED_LOG_TYPE) as SelectState) ?? undefined,
```
khelif96, line 28: Why did you make this change it seems to do the same thing but is arguably slightly less readable.

src/pages/LogView/LoadingPage/useResolveLogURLAndRenderingType.ts
```diff
@@ -194,12 +199,19 @@ export const useResolveLogURL = ({
     default:
       break;
   }
+
+  // renderingType is only propagated for evergreen test logs as of DEVPROD-1435.
+  const renderingType =
+    (testData?.task?.tests.testResults[0]?.logs
```
khelif96, line 205: `renderingType` will default to `default` for resmoke logs which is confusing. I think we should try to be consistent about how this value is treated and used.

khelif96 (changes requested):
I think this can be done in a way that is a little more flexible for expanding to different rendering types. Right now the assumption is that there are only 2 types of logs and doesn't leave room for future log types like how the previous implementation did. 

If we are going to be using `renderingType` to determine log render behavior we should try to be consistent across the app about how it is used instead of having it only apply to a very specific subset of test logs.

src/pages/LogDrop/ParseLogSelect/index.tsx
SupaJoon: `LogTypes` includes values that are unrelated to the Select component state which makes it harder to infer usage and less readable. We should be more strict with our typings and exclude unused values.

src/components/LogRow/RowRenderer/index.tsx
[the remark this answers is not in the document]
SupaJoon: I think this logic doesn't make sense because it doesn't do a good job at holistically bringing in `renderingType` to Parsley.  The logic that you reviewed utilizes parsley route and renderingType only for logs. I just pushed some code to decouple the route from renderingType, utilizes testlog renderingType from the GQL endpoint but also extrapolates rendering type per logType.

src/pages/LogDrop/FileDropper.tsx
```diff
@@ -42,7 +42,14 @@ const FileDropper: React.FC = () => {
...
     (logType: LogTypes | undefined) => {
       if (logType) {
-        setLogMetadata({ isUploadedLog: true, logType });
+        setLogMetadata({
+          isUploadedLog: true,
+          logType,
+          renderingType:
+            logType === LogTypes.RESMOKE_LOGS
+              ? SupportedLogRenderingTypes.Resmoke
+              : SupportedLogRenderingTypes.Default,
```
khelif96, line 51: Doing the above we can also avoid this.

src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -38,6 +40,24 @@ const ParsleyRow: RowRendererFunction = ({ logType, processedLogLines }) => {
       ? new RegExp(`${highlights.map((h) => `(${h})`).join("|")}`, "gi")
       : undefined;
 
+  let Row: React.FC<ResmokeRowProps> | React.FC<AnsiRowProps>;
```
khelif96, line 43: This is more flexible and avoids exporting the the Row props.
```suggestion
  let Row: typeof ResmokeRow | typeof AnsiRow;
```

khelif96 (changes requested):

src/pages/LogDrop/FileDropper.tsx
khelif96: hmm i think this may be ok, Since this could be used to distinguish between log types for log sectioning project? I do think it would be important to iron that out in your project or at some point before it.

src/pages/LogView/LoadingPage/useResolveLogURLAndRenderingType.ts
```diff
@@ -169,13 +176,18 @@ export const useResolveLogURL = ({
...
     case LogTypes.EVERGREEN_TEST_LOGS: {
       if (!taskID || !execution || !testID || isLoadingTest) {
         break;
       }
-      const { url, urlRaw } = testData?.task?.tests.testResults[0]?.logs || {};
+      const {
+        renderingType: renderingTypeFromQuery,
+        url,
+        urlRaw,
+      } = testData?.task?.tests.testResults[0]?.logs || {};
```
khelif96, line 190: Can we leave a comment here. 
Something simple such as if we fetched test results use that URL otherwise fallback to hardcoded routes?

src/pages/LogView/LoadingPage/useResolveLogURLAndRenderingType.ts
```diff
@@ -108,6 +111,7 @@ export const useResolveLogURL = ({
   let htmlLogURL = "";
   let jobLogsURL = "";
   let legacyJobLogsURL = "";
+  let renderingType = "";
```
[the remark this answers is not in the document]
khelif96: This is still necessary. Otherwise the type alias on line 206 has no effect.

src/pages/LogView/LoadingPage/useResolveLogURLAndRenderingType.ts
```diff
@@ -189,17 +201,32 @@ export const useResolveLogURL = ({
           text: false,
         });
       downloadURL = rawLogURL;
+      if ((renderingTypeFromQuery || "") in LogRenderingTypes) {
+        renderingType = renderingTypeFromQuery as LogRenderingTypes;
+      } else if (!renderingTypeFromQuery) {
+        renderingType = LogRenderingTypes.Default;
+      } else {
+        renderingType = LogRenderingTypes.Default;
```
khelif96, line 209: I think this can be refactored out. The first if statement handles the case where renderingTypeFromQuery is undefined. `renderingTypeFromQuery` can't be in `LogRenderingTypes`

khelif96: If we assign a default value for `renderingType` like my suggestion on on L115 this assignment in the else statement becomes unnecessary as well

src/pages/LogView/LoadingPage/useResolveLogURLAndRenderingType.ts
```diff
@@ -189,17 +201,32 @@ export const useResolveLogURL = ({
...
+        reportError(new Error("Encountered unsupported renderingType"), {
+          context: {
+            rawLogURL,
+            unsupportedRenderingType: renderingTypeFromQuery,
+          },
+        }).warning();
```
khelif96, line 215: This should probably be logged at `.severe()` since this indicates we encountered some sort of unsupported rendering type.

khelif96 (commented):
nit. Can you capitalize DEVPROD in the pr title

Great job just a few more comments and follow up from my last review.

src/pages/LogDrop/FileDropper.tsx
```diff
@@ -40,9 +40,17 @@ const FileDropper: React.FC = () => {
...
   const onParse = useCallback(
-    (logType: LogTypes | undefined) => {
-      if (logType) {
-        setLogMetadata({ isUploadedLog: true, logType });
+    (renderingType: LogRenderingTypes | undefined) => {
+      if (renderingType) {
+        const logType =
+          renderingType === LogRenderingTypes.Resmoke
+            ? LogTypes.RESMOKE_LOGS
+            : LogTypes.EVERGREEN_TASK_LOGS;
```
khelif96, line 48: Once sectioning for task logs is enabled we may want to change this.

src/pages/LogView/LoadingPage/useResolveLogURLAndRenderingType.ts
khelif96: I may be missing what you mean but Afaict `renderingType` doesn't get overridden anywhere else so I still think it's not necessary. Can you point me to where it does?

khelif96 (approved):
Great job on this! 🚀  mod my comment

PR #493, opened 2024-02-21
DEVPROD-4976: Use ec2.assume_role command

minnakt:
DEVPROD-4976

### Description 
Use `ec2.assume_role` and its associated credentials in `.evergreen.yml`.

### Testing 
- The tests should pass and successfully upload files to S3

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -19,12 +19,18 @@ modules:
 #            Parsley Functions        #
 #######################################
 functions:
+  assume-ec2-role:
+    command: ec2.assume_role
+    params:
+      role_arn: ${ASSUME_ROLE_ARN}
+
   attach-codegen-diff:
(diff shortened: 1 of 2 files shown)


.evergreen.yml
```diff
@@ -341,14 +355,16 @@ functions:
     params:
       working_dir: parsley
       shell: bash
+      env:
```
julianedwards, line 358: Non-blocking suggestion: would it be better to use `subprocess.exec` instead of `shell.exec` throughout this config?

julianedwards (approved):

.evergreen.yml
minnakt: for sure! I fixed one of the easy ones, but the other ones make me a bit nervous so I'll open a ticket to audit this usage!

PR #509, opened 2024-03-13
DEVPROD-5130: Case on renderingType during log ingestion, introduce LOCAL_UPLOAD logType and fix bug related to exposing renderingType from hook

SupaJoon:
DEVPROD-5130

### Description 
These code changes:

- Introduce LogTypes.LOCAL_UPLOAD
- Fix useResolveLogURLAndRenderingType to correctly return the `resmoke` rendering type when applicable
- Update some components to read LogType from context instead of props
- Update the breadcrumb component to show the test name for EVG test logs. 
- Makes some changes to how logs from S3 are fetched

### Testing 
I unit tested the `useResolveLogURLAndRenderingType` hook and then asserted the correct rendering logic is utilized in E2E tests. It's probably useful to ensure ingestLogs is called with the correct rendering type in context which means we would have to unit test LoadingPage. We should actually do this eventually because LoadingPage makes guarantees about the state of context during the log loading phase of the app (DEVPROD-5891).


sophstad (commented):
Any chance you could open this in the new repo? The Evergreen YAML changes in particular may be a bit different and I think it would be a lot easier to centralize the review process 🙏 let me know if I can help at all!

SupaJoon:
Ported this PR to evergreen-ci/ui

PR #514, opened 2024-03-18
DEVPROD-5529 Update logkeeper bootstrap script with additional environment variables

khelif96:
DEVPROD-5529

### Description 
This should hopefully make starting logkeeper for the first time easier.

### DIFF
--- a/scripts/bootstrap-logkeeper.sh
+++ b/scripts/bootstrap-logkeeper.sh
@@ -49,4 +49,4 @@ fi
 
 
 echo "Use the following command to start logkeeper:"
-echo "${YELLOW}LK_CORS_ORIGINS=http:\/\/localhost:\\\d+ go run main/logkeeper.go --localPath $PWD/bin/_bucketdata${NC}"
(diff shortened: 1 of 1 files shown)


scripts/bootstrap-logkeeper.sh
```diff
@@ -49,4 +49,4 @@ fi
```
minnakt, line 50: i notice a random unused variable called `SPRUCE_DIR` in here 👀 could maybe delete

scripts/bootstrap-logkeeper.sh
```diff
@@ -49,4 +49,4 @@ fi
 
 
 echo "Use the following command to start logkeeper:"
```
minnakt, line 51: it could be helpful to mention that the command should be run from your logkeeper directory (wherever it exists)

minnakt (approved):

scripts/bootstrap-logkeeper.sh
khelif96: Good eye!

PR #515, opened 2024-03-18
DEVPROD-5665 Update cypress e2e task to not report to cypress cloud

khelif96:
DEVPROD-5665

### Description 
irt the slack thread we will no longer be sending data to cypress cloud. 

### Testing 
The associated task still displays all of the necessary logs and links to files in the files tab

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -277,7 +277,12 @@ functions:
       shell: bash
       script: |
         ${PREPARE_SHELL}
-        yarn cy:run --reporter junit
+        # Allow spec filtering for an intentional patch.
+        if [[ "${requester}" == "patch" ]]; then
+          yarn cy:run --reporter junit --spec "${cypress_spec}"
+        else
+          yarn cy:run --reporter junit
+        fi
 
 
   yarn-eslint:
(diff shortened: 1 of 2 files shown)


.evergreen.yml
```diff
@@ -277,13 +277,14 @@ functions:
       shell: bash
       script: |
         ${PREPARE_SHELL}
-        # Only record to cypress cloud if this is a pr or a mainline commit.
-        if [[ "${requester}" == "github_pr" || "${requester}" == "commit" || "$requester" == "patch" ]]; then
-          yarn cy:run --record --key "${parsley_cypress_record_key}" --reporter junit
+        # Allow spec filtering for an intentional patch.
+        if [[ "${requester}" == "patch" ]]; then
+          yarn cy:run --reporter junit --spec "${cypress_spec}"
```
minnakt, line 282: i think this already works without it, but should we add the `parameters` block like we have in spruce? I'm not really sure what benefit the block gives though 😅

minnakt (approved):

.evergreen.yml
khelif96: Sure just added it for consistency! 
I was kind of curious about what it did too since I didn't see it in the docs.
I think it does 2 things; applies default values for patch params that are loaded in as expansions and also provides a description when using the evergreen list command in the cli.

